PR #114, opened 2015-03-16
packages: add option to suppress warning messages printed directly to console

jbohren:


wjwwood:
The changes look reasonable, but there is a reason the warnings are being printed... For building locally, I think this is a valid scenario for ignoring warnings like this.

It would be nice to get the warnings rather than only having them printed to the screen or completely avoided. Maybe rather than `suppress_warning` you could pass `warnings`, which defaults to `None`. If it is `None` then messages are printed, but if it is a `list` then the warnings are appended to the list but not printed.

src/catkin_pkg/package.py
```diff
@@ -357,18 +359,19 @@ def parse_package(path):
 
     with open(filename, 'r') as f:
         try:
-            return parse_package_string(f.read(), filename)
+            return parse_package_string(f.read(), filename, suppress_warnings)
```
dirk-thomas: Please pass kwargs as kwargs instead of positional. Same below.

dirk-thomas:
Very good point from William, changing the new parameter to "collect" the warnings would be even better.

jbohren:
> It would be nice to get the warnings rather than only having them printed to the screen or completely avoided. Maybe rather than suppress_warning you could pass warnings, which defaults to None. If it is None then messages are printed, but if it is a list then the warnings are appended to the list but not printed.

Yeah, I'm down.

jbohren:
Ok. c9dfa9b adds warnings to a list passed into the `warnings` param if it isn't set to `None`

wjwwood:
Other than the comment, lgtm.

src/catkin_pkg/package.py
```diff
@@ -226,10 +229,16 @@ def validate(self):
...
-                print(('WARNING: Metapackage "%s" should not have other dependencies besides a ' +
-                       'buildtool_depend on catkin and run_depends.') % self.name, file=sys.stderr)
+                new_warnings.append('Metapackage "%s" should not have other dependencies besides a '
+                                    'buildtool_depend on catkin and run_depends.' % self.name)
+
+        for warning in new_warnings:
+            if warnings is None:
+                print('WARNING: ' + warning, file=sys.stderr)
+            elif warning not in warnings:
+                warnings += warning
```
dirk-thomas: If `warnings` is a list and `warning` is a string than this operation won't do the right thing?

dirk-thomas:
Can you please add a unit test for this new feature to ensure it works as expected.

jbohren:
Yeah I'll add a test.

jbohren:
Test added.

test/test_metapackage.py
```diff
@@ -76,6 +86,14 @@ def test_validate_metapackage(self):
                 else:
                     _validate_metapackage(path, package)
 
+    def test_collect_warnings(self):
+        """Tests warnings collection"""
+        warnings = []
+        pkgs_dict = find_packages(test_data_dir, warnings=warnings)
+
+        for w, we in zip(warnings, test_expected_warnings):
+            assert(w == we)
```
wjwwood: You should try to give a second argument to `assert` when ever you use it, because this line is failing on the farm, but I don't know why exactly.

You can also just do `self.assertEqual(sorted(w), sorted(we))`, or if you are certain what the order should be `self.assertEqual(w, we)`. The `self.assertEqual` will automatically give you a nice comparison out if it fails.

jbohren: Yeah, I'll get it working on the farm.

wjwwood:
lgtm, @dirk-thomas?

src/catkin_pkg/package.py
```diff
@@ -357,18 +367,19 @@ def parse_package(path):
 
     with open(filename, 'r') as f:
         try:
-            return parse_package_string(f.read(), filename)
+            return parse_package_string(f.read(), filename, warnings)
```
dirk-thomas: Please always pass kwargs with the explicit keyword instead of relying on the position. Same below.

test/test_metapackage.py
```diff
@@ -26,9 +26,19 @@
     'no_buildtool_depend_catkin': [InvalidMetapackage, 'No buildtool dependency on catkin', None],
     'no_cmake': [InvalidMetapackage, 'No CMakeLists.txt', None],
     'no_metapackage_tag': [InvalidMetapackage, 'No <metapackage/> tag in <export>', None],
+    'NonConformingName': [None, None, None],
     'valid_metapackage': [None, None, None]
 }
 
+test_expected_warnings = [
+    'Metapackage "invalid_depends" should not have other dependencies besides a '
```
dirk-thomas: Since you wrapped everything else to fit in 80 columns: the `a ` exceed the line length and should be wrapped to the next line.

test/test_metapackage.py
```diff
@@ -76,6 +86,16 @@ def test_validate_metapackage(self):
...
 
+    def test_collect_warnings(self):
+        """Tests warnings collection"""
+        warnings = []
+        pkgs_dict = find_packages(test_data_dir, warnings=warnings)
+
+        warnings.sort()
+        test_expected_warnings.sort()
+        for w, we in zip(warnings, test_expected_warnings):
+            assert w == we, 'Warning message "%s" != "%s"' % (w,we)
```
dirk-thomas: The assertion could get really weird as soon as the two lists are not aligned anymore. I would recommend to simply compare two sets with a single assert:

    assert set(warnings) == set(test_expected_warnings), 'Unexpected difference in warnings: %s != %s' % (warnings, test_expected_warnings)

wjwwood: But by that logic he should do sorted lists, or additionally assert that the set is equal to the original list. Comparing sets alone will not catch one list having duplicates that the other does not.

I still think my previous comment applies:

https://github.com/ros-infrastructure/catkin_pkg/pull/114#discussion_r26631509

I'd recommend using `self.assertEqual` on either the lists or the sorted version of the lists, because if one warning doesn't match, it will most likely require the context of the list of warnings to figure out what is the issue.

jbohren: Is there anything wrong with the current approach here of comparing two sorted lists?

wjwwood: I think it's ok as is, but the `self.assertEqual` should be used in the future, since it gives better feed back on the error when it occurs.

dirk-thomas: Asserting on the identity of the items will result in a not very helpful error message. Instead please just assert that both sorted lists are equal.

jbohren:
Ready for review again.

src/catkin_pkg/package.py
```diff
@@ -529,7 +540,7 @@ def parse_package_string(data, filename=None):
     if errors:
         raise InvalidPackage('Error(s) in %s:%s' % (filename, ''.join(['\n- %s' % e for e in errors])))
 
-    pkg.validate()
+    pkg.validate(warnings)
```
dirk-thomas: Last comment - promise :wink: 
This one should use the keyword too: `warnings=warnings`.

jbohren:
Ready again again.

dirk-thomas:
Thank you for making all the changes for the numerous small comments. LGTM.

PR #117, opened 2015-04-03
optimization: avoid seemingly unnecessary deepcopies

jbohren:
I noticed that about 15% of a call to `catkin build --dry-run` was just doing `copy.deepcopy` in `catkin_pkg`. Is there a reason why these need to be deep copies?

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -37,7 +37,7 @@ representation.
 
 from __future__ import print_function
 
-from copy import deepcopy
+from copy import copy
 import os
 import re
 import sys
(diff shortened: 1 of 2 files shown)


dirk-thomas:
In all but one cases `deepcopy` is used is to ensure that a dependency instance is not  partly reused when being inserted into multiple types. Since the `Dependency` class is not nested and only contains slots with strings I would not even expect `deepcopy` to be more expensive than `copy`.

From the docs: "The difference between shallow and deep copying is only relevant for compound objects (objects that contain other objects, like lists or class instances)" (https://docs.python.org/2/library/copy.html).

Are you sure that these uses of `deepcopy` are actually the problem?

The only other case were `deepcopy` is for cloning the passed packages in the `_sort_decorated_packages` function. I could see why this might take more time but the function should only be called once.

dirk-thomas:
Can you please provide more information on this or a reproducible example?

dirk-thomas:
I measured the runtime of invoking `catkin_pkg.packages.find_packages_allowing_duplicates` which found roughly 1k of packages. The time was exactly the same with as without this patch.

Without further feedback on this I will close this ticket. If there is more information provided the ticket can be reopened.

Issue #79, opened 2013-11-18
catkin_create_pkg could hint at _EXPORTED_TARGETS


dirk-thomas:
Moved from ros/catkin#548 filled by @tkruse:

The CMakeLists.txt created by catkin_create_pkg contains a lot of help and templates to guide users into writing clean catkin packages.

I suggest adding a small block to encourage users to use the ${catkin_EXPORTED_TARGETS} variable in add_dependencies when using e.g. messages/dynamic_reconfigure from other packages. The help text should IMO explicitly mention message generation and dynamic_reconfigure. possibly it could mention build performance effects, though I am not in favor of that.

PR #119, opened 2015-05-30
add dynamic reconfigure example and use exported targets in generated CMake template

dirk-thomas:
This adds example code for `dynamic_reconfigure`. @esteve Please check that the example code and instructions are correct.

Additionally the CMake template uses the exported targets rather then the `_generate_messages_cpp` target to address #79.

@esteve @tfoote @wjwwood Please review.


tfoote:
+1

wjwwood:
+1

esteve:
+1

Issue #55, opened 2013-07-15
catkin_generate_changelog should escape words ending with underscore


davetcoleman:
I automatically generated change logs for my packages using my git log using:
```
catkin_generate_changelog
```
Then I prepare the release and I get error messages like:
```
<string>:21: (ERROR/3) Unknown target name: "plugin_loaded".
```
After discussing this issue with @wjwwood in this issue: https://github.com/ros-infrastructure/bloom/issues/189 I learned the error was because I had git commits saying things like
```
* gazebo_ros_api_plugin: set plugin_loaded_ flag to true in
```
Where ``plugin_loaded_`` is a member variable that I am discussing. 

This will throw an error with the RST Changelog format because RST treats those as [link targets](http://docutils.sourceforge.net/docs/user/rst/quickstart.html#sections). 

I propose that ``catkin_generate_changelog`` automatically escape these words to read, for example:
```
* gazebo_ros_api_plugin: set ``plugin_loaded_`` flag to true in
```
This seems to be the most intuitive default behavior because it won't generate unexpected errors without good explanations. In the meantime I have [documented](http://www.ros.org/wiki/bloom/Tutorials/ReleaseCatkinPackage#preview) this issue in the bloom release tutorial.

davetcoleman:
Thanks!

PR #120, opened 2015-05-30
escape trailing underscores when generating changelog

dirk-thomas:
Addresses #55.

Since the patch does not perform any parsing please consider any cases where this might fail.

@esteve @tfoote @wjwwood Please review.

### DIFF
--- a/src/catkin_pkg/changelog_generator.py
+++ b/src/catkin_pkg/changelog_generator.py
@@ -257,6 +257,7 @@ def generate_version_content(log_entries, vcs_client=None, skip_contributors=Fal
         lines = msg.splitlines()
         lines = [l.strip() for l in lines]
         lines = [l for l in lines if l]
+        lines = [escape_trailing_underscores(l) for l in lines]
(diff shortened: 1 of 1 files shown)


tfoote:
This looks reasonable. +1

wjwwood:
That looks ok to me too, I had to lookup some of the regex syntax, but it makes sense. +1

src/catkin_pkg/changelog_generator.py
```diff
@@ -266,6 +267,13 @@ def generate_version_content(log_entries, vcs_client=None, skip_contributors=Fal
     return data
 
 
+def escape_trailing_underscores(line):
+    if line.endswith('_'):
+        line = line[:-1] + '\_'
+    line = re.sub(r'(\w+)_([^\w])', '\\1\\_\\2', line)
```
esteve: Please add a comment explaining what the regex does, it's obvious now but may not be six months from now (or even a week :-))

esteve:
+1

src/catkin_pkg/changelog_generator.py
dirk-thomas: Done.

PR #122, opened 2015-06-26
add module for reading and writing built by markers

wjwwood:
See: https://github.com/ros/catkin/issues/742

### DIFF
--- a/test/test_tool_detection.py
+++ b/test/test_tool_detection.py
@@ -1,0 +1,18 @@
+import os
+
+from catkin_pkg.tool_detection import get_previous_tool_used_on_the_space
+from catkin_pkg.tool_detection import mark_space_as_built_by
+
+from .util import in_temporary_directory
+
+
+@in_temporary_directory
+def test_get_previous_tool_used_on_the_space():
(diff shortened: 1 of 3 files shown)


tfoote:
+1

jbohren:
+1

test/test_tool_detection.py
```diff
@@ -0,0 +1,18 @@
...
+from catkin_pkg.tool_detection import get_previous_tool_used_on_the_space
+from catkin_pkg.tool_detection import mark_space_as_built_by
+
+from .util import in_temporary_directory
+
+
+@in_temporary_directory
+def test_get_previous_tool_used_on_the_space():
+    res = get_previous_tool_used_on_the_space('folder_that_does_not_exist')
+    assert(res is None, res)
```
dirk-thomas: `assert` is not a function but a statement. This check will always be true no matter what `res` is.

Same below.

wjwwood: Good catch, too much C lately...

https://github.com/ros-infrastructure/catkin_pkg/commit/e9b73902a1164ad7f0eb2bb16b358cef2f2300c7

dirk-thomas:
lgtm

PR #123, opened 2015-07-01
Add LICENSE.txt

cottsay:
Fedora packaging guidelines recommend including a standalone license file with packaged software. This was taken from one of the python files' headers.

Thanks,

--scott


wjwwood:
Is the `.txt` extension necessary. In our other repositories we usually just use `LICENSE`.

cottsay:
No problem - I'll remove it.

cottsay:
Fixed - sorry for the delay.

Issue #96, opened 2014-06-25
catkin_create_pkg: start using package.xml format 2


jack-oquin:

scpeters:
I have a work in progress to fix this in #191.

PR #126, opened 2015-08-06
Proposal: Use empy for template substitution

ipa-mdl:
The current implementation uses string.Templates to create new package scaffolds.
I would like to implement --brief (name TBD) argument to catkin_create_pkg that generates shorter files with fewer explanations.
However, string.Templates does not support conditional expansion, so I migrated the code to empy.
As an alternative different template sets could be switched with a to-be-implemented --template argument.

We are planning to provide package manifest format two templates (#96).
In addition the templates should include hints based on the dependencies, empy is more suitable for this.

If you do not want catkin_pkg to depend on empy, we will see if we can implement this otherwise or if we have to create a new tool for our purpose.


dirk-thomas:
These templates are also used by CMake code within `catkin` (e.g. https://github.com/ros/catkin/blob/a09f96766780ab1190ef1fa61d1bf9b406089c7e/cmake/catkin_workspace.cmake#L61-L63) using the `configure_file` CMake function.

Since `catkin_pkg` is used across all ROS distros but `catkin` is a distro specific package it will be pretty difficult to make such a change without break something.

Therefore I don't think it is feasible to change the template type.

ipa-mdl:
Okay, that's a good point!
We will at least update the templates to format 2.

For the convenience features I will have a look at https://github.com/DLu/roscompile

PR #128, opened 2015-08-20
Replacing @name with ${PROJECT_NAME}

aleksandaratanasov:
Replacing almost all occurrences of @name with ${PROJECT_NAME} in order to make changes of the package name easier to handle. The only places (excluding the project's name line at the top of the CMakeLists.txt) where @name is still used are the source files. This will prevent changes to name of project affect the default source files defined in the CMakeLists.txt.


dirk-thomas:
Thanks.

PR #129, opened 2015-10-31
Fix KeyError in os.environ['CATKIN_PREFIX_PATH']

mylxiaoyi:
Fix KeyError in os.environ['CATKIN_PREFIX_PATH'] under both python2.7 and python3.4

### DIFF
--- a/src/catkin_pkg/workspaces.py
+++ b/src/catkin_pkg/workspaces.py
@@ -50,7 +50,7 @@ def get_spaces(paths=None):
     if paths is None:
         if 'CMAKE_PREFIX_PATH' not in os.environ:
             raise RuntimeError('Neither the environment variable CMAKE_PREFIX_PATH is set nor was a list of paths passed.')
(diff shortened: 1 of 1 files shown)


src/catkin_pkg/workspaces.py
```diff
@@ -50,7 +50,7 @@ def get_spaces(paths=None):
     if paths is None:
         if 'CMAKE_PREFIX_PATH' not in os.environ:
             raise RuntimeError('Neither the environment variable CMAKE_PREFIX_PATH is set nor was a list of paths passed.')
-        paths = os.environ['CMAKE_PREFIX_PATH'].split(os.pathsep) if os.environ['CMAKE_PREFIX_PATH'] else []
+        paths = os.environ['CMAKE_PREFIX_PATH'].split(os.pathsep) if 'CMAKE_PREFIX_PATH' in os.environ else []
```
dirk-thomas: This won't work correctly if the environment variable is set but empty.

dirk-thomas:
The same check is already being performed two lines above. Therefore I don't see how you could run into the `KeyError` in the line you patched.

I will close this for now but please feel free to provide more detailed information if this is still a problem.

PR #131, opened 2015-11-20
Fixes #130.

### LINKED ISSUE #130: catkin_generate_changelog --all Includes Non-Ancestor Tags
When generating a changelog, "future" versions (and tags on unmerged branches) should not be included in the changelog. The current behavior of `catkin_generate_changelog --all` is to generate a changelog entry for every tag in the git repository, even if that tag is not an "ancestor" of the current `HEAD`.

Proposed fix:
Use some parsing of `git log --simplify-by-decoration --pretty="%d" HEAD` instead of `git tag` to generate the tag list for changelogs. This will only show tags that are "ancestors" of the current `HEAD`.

This doesn't address other VCSs, but I think git is the main usecase for `catkin_generate_changelog`.

evenator:
When generating a changelog, only tags that are ancestors of the current HEAD
should be included in the changelog. The previous behavior of
`catkin_generate_changelogs --all` was to make changelog entries for every tag
in the repository.

This commit changes the behavior of GitClient.get_tags() such that only
ancestors of the current HEAD are returned.

### DIFF
--- a/src/catkin_pkg/changelog_generator_vcs.py
+++ b/src/catkin_pkg/changelog_generator_vcs.py
@@ -146,11 +146,11 @@ class GitClient(VcsClientBase):
         return result['output']
 
     def get_tags(self):
-        cmd_tag = [self._executable, 'tag']
+        cmd_tag = [self._executable, 'log', '--simplify-by-decoration', '--pretty="%d"']
         result_tag = self._run_command(cmd_tag)
         if result_tag['returncode']:
             raise RuntimeError('Could not fetch tags:\n%s' % result_tag['output'])
-        tag_names = result_tag['output'].splitlines()
+        tag_names = re.findall("(?<=tag: )([^)]+)", result_tag['output'])
 
         tags = []
         for tag_name in tag_names:
--- a/src/catkin_pkg/changelog_generator_vcs.py
+++ b/src/catkin_pkg/changelog_generator_vcs.py
@@ -146,11 +146,15 @@ class GitClient(VcsClientBase):
         return result['output']
 
     def get_tags(self):
-        cmd_tag = [self._executable, 'log', '--simplify-by-decoration', '--pretty="%d"']
+        # Get a decorated log, use the refnames to find the ancestor tags
+        cmd_tag = [self._executable, 'log', '--simplify-by-decoration', '--decorate', '--pretty=oneline']
         result_tag = self._run_command(cmd_tag)
         if result_tag['returncode']:
             raise RuntimeError('Could not fetch tags:\n%s' % result_tag['output'])
-        tag_names = re.findall('tag: ([^),]+)[),]', result_tag['output'])
+        # Parse a comma-separated list of refname decorators out of the log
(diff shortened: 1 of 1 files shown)


src/catkin_pkg/changelog_generator_vcs.py
```diff
@@ -146,11 +146,11 @@ def _get_author(self, hash_):
         return result['output']
 
     def get_tags(self):
-        cmd_tag = [self._executable, 'tag']
+        cmd_tag = [self._executable, 'log', '--simplify-by-decoration', '--pretty="%d"']
```
dirk-thomas: Since the output contains more than tag names I am worried about parts of it being locale specific. Can you try it with a different locale and check if `tag` is still called `tag`? If not, the command needs to be run with a fixed locale like `LANG=en_US.UTF-8`.

The double quotes around `%d` are not necessary.

src/catkin_pkg/changelog_generator_vcs.py
```diff
@@ -146,11 +146,11 @@ def _get_author(self, hash_):
...
         result_tag = self._run_command(cmd_tag)
         if result_tag['returncode']:
             raise RuntimeError('Could not fetch tags:\n%s' % result_tag['output'])
-        tag_names = result_tag['output'].splitlines()
+        tag_names = re.findall("(?<=tag: )([^)]+)", result_tag['output'])
```
dirk-thomas: The regular expression uses a lookbehind which seems a bit too complicated to me. Can you please simply the regex to not require that.

The regex could also be made more specific by ensuring that the tag name:
* is preceeded by either `(` or `\s` and
* is followed by either `,` or `)`.

Please use single quotes instead of double quotes as in the rest of the file.

evenator:
I just discovered that git 1.7.9 (which is packaged with Ubuntu Precise) implements the %d flag differently (it doesn't prepend the ref name with the type). I'm not sure when the change was made, but this works in git 1.9.1 (packaged with Ubuntu Trusty).

git 1.7.9 does prepend the ref types when using `git log --simplify-by-decoration --decorate --pretty="oneline"`, but that also output the commit hash and brief, so more parsing would be required.

evenator:
If git 1.7 support is desired, this commit changes the git command to work in 1.7 as well as 1.9:
https://github.com/evenator/catkin_pkg/commit/0a36f6c19ab5a12bd409174c5b0d766592935b9e

Let me know if I should merge it in.

dirk-thomas:
Yes, please add your change to support older git versions.

It looks like the regex would also match for lines which only contain "tag: something" as a commit message though.

evenator:
That pair of regexes would erroneously detect a tag in a commit that:

* Is neither a tag nor a branch nor HEAD (i.e. has no decorations)
* Is still printed by git log --simplify-by-decoration
* Has a commit message that begins with text that contains tag: something within parentheses

The first two conditions seem to limit this to the root of the tree and certain merge commits. I don't think I can completely eliminate the possibility of a spurious tag detection, but I can it make it less likely by making the regex that extracts the decorations stricter.

dirk-thomas:
Thanks. Looks and works well for me. I squashed and cherry-picked it in 58eb59e2632baf6591b788cc4f45ab06e1f2b947.

PR #139, opened 2016-04-14
stop using distutils

gerkey:


wjwwood:
Looks reasonable to me.

gerkey:
I'm new in to the ros-infrastructure club. Anything that I should test or check before merging?

wjwwood:
I'd just let @dirk-thomas have a look, since he maintains this one usually.

dirk-thomas:
LGTM

Just out of curiosity: in which case was the relative path a problem?

wjwwood:
I don't exactly know. I think it might have had to do with the weird thing we do where ament_tools will symlink the setup.py and some other stuff to the build folder to prevent build artifacts from getting into the source space. I'm not sure exactly under what conditions this becomes a problem, i.e. `__file__` is relative.

dirk-thomas:
If only the setup file is symlinked I don't see how `abspath` (without `realpath`) would make a difference in that case.

gerkey:
I'd just tried the use case we're working on and, indeed, the use of `abspath` doesn't make any difference. It was something that we tried along the way and thought was having an effect, but it apparently doesn't. The key change for our use case is to not do a symlink install at all (ticket forthcoming on that issue).

I undid the `abspath` change and squashed. OK to merge?

dirk-thomas:
+1

Issue #140, opened 2016-04-21
Is following version convention critical? (warnings of ompl version style)


wkentaro:
I have a lot of below warnings when using ROS packages, and it seems this problem remains for a long time.
https://github.com/ros-gbp/ompl-release/releases
```
WARNING: Package "ompl" does not follow the version conventions. It should not contain leading zeros (unless the number is 0).
```

Is following version convention crucial? (If so why ompl can be released in ROS farm?)
If not, I'd like to remove the line in https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/package.py#L188
for few warnings.

Related to: https://github.com/ros-gbp/ompl-release/issues/4

davetcoleman:
+1, its pretty annoying getting this warning. ive been getting it for months but am finally investigating. its because the package.xml says ``<version>1.0.0003094</version>`` instead of ``<version>1.0.3094</version>``

dirk-thomas:
The package in question `ompl` has last been released in 2014 into Indigo. The Jade version is basically the same. The package could easily be rereleased with a compliant version number. I don't see a reason to remove the check / warning in `catkin_pkg` just because some package decides to not follow the convention.

mamoll:
The leading zeros are @isucan's idea. The version number can be broken down as:
```
<major>.<minor>.<patchlevel>00<hgrevision>
```
In the beginning when the OMPL package changed rapidly this made sense. I'd say that OMPL has been very stable. The ROS deb packages also don't track hg revisions and (should) track official releases. So in my opinion the 0's and the hg revision part of the version number can be dropped. It'd be good if the OMPL package in ROS could be updated to version 1.1, though.

davetcoleman:
Is that format followed manually or is it coded somewhere? Could you change it?

ROS Kinetic would be ideal for OMPL 1.1, though we could sneak it into Jade too, probably. 1.1 does not include the C++11 support, correct?

mamoll:
I think it is just hard-coded here: https://github.com/ros-gbp/ompl-release/blob/master/tracks.yaml.
C+11 support is currently unreleased.

wjwwood:
It would be great to see a released version of C++11 support (since the C++11 compiler is now a minimum requirement for Kinetic). However, if given the choice between C++11 support on an unreleased branch and a released version without, I'd favor the released version for Kinetic since it will be an LTS. Maybe interested parties could lean on the OMPL maintainers to get another release out.

[...]

PR #141, opened 2016-04-28
Remove picky warning about leading zeros

davetcoleman:
OMPL uses ``<version>1.0.0003094</version>`` which seems like it should be ok but we always get warnings

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -184,8 +184,6 @@ class Package(object):
             errors.append('Package version must not be empty')
         elif not re.match('^[0-9]+\.[0-9]+\.[0-9]+$', self.version):
             errors.append('Package version "%s" does not follow version conventions' % self.version)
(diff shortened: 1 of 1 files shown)


davetcoleman:
Closes https://github.com/ros-infrastructure/catkin_pkg/issues/140

davetcoleman:
Closes https://github.com/ros-gbp/ompl-release/issues/4

dirk-thomas:
`ompl` is currently the only package I am aware of which violates this rule. It hasn't been modified for almost two years and a simple rerelease of the package with a compliant version number would resolve this. Therefore I don't see any argument why this warning should be removed.

wjwwood:
The whole point of the warning is to discourage non-compliant version numbers. Also -1 to relaxing the warning. It seems like just going with a released version is the right thing to do: https://github.com/ros-infrastructure/catkin_pkg/issues/140#issuecomment-215554011

davetcoleman:
no problem, thanks

PR #143, opened 2016-05-05
Always read package files as utf8 on python3.

de-vri-es:
This PR forces files to be treated as UTF-8 on python3. This avoids breaking down on non-utf8 locales for `package.xml` files containing unicode.

I personally encountered this when building in a chroot. While I arguably should fix the locale of the chroot, it is not good to depend on the locale for properly parsing `package.xml` files.

This PR should fix ros-infrastructure/rosdep#309

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -365,7 +365,13 @@ def parse_package(path, warnings=None):
     else:
         raise IOError('Path "%s" is neither a directory containing a "%s" file nor a file' % (path, PACKAGE_MANIFEST_FILENAME))
 
-    with open(filename, 'r') as f:
+    # Force utf8 encoding for python3.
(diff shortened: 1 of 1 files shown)


de-vri-es:
Note that python2 doesn't support the `encoding` argument for the `open()` function, otherwise it would have made sense to always pass `encoding='utf8'` to the `open()`.

de-vri-es:
ping?

dirk-thomas:
This looks good to me.

Just as a comment for you case: you still need to set the locale in your chroot environment. Otherwise when a script e.g. prints the maintainer name it parsed out of a manifest it will fail the same way if the string contains non-ascii characters..

de-vri-es:
True, I'll get round to that some time ;)

Thanks for the merge.

PR #144, opened 2016-05-10
Replace scripts by setuptools console entry points

dmerejkowsky:
This makes it possible (among other things) to use `find_executable(catking_find_pkg)` on Windows and finds a .exe file that works.

Without this patch, you'll find the "shebang" script and it's likely to not work on Windows.

Downside is that you have to run `setup.py develop` before being able to run the scripts.

Upside is that you don't need hacks to tests the scripts:
https://github.com/ros-infrastructure/catkin_pkg/compare/master...dmerejkowsky:console-entry-points?expand=1#diff-e70daccfe471bb3da351649e55210e0fL12


dirk-thomas:
The current patch is very difficult to review since it is unclear which code has just been moved and where modifications have happened. Also putting the content of all scripts into a single module seems to be a bad idea since it makes it less modular and clear what each script uses.

Can you please refactor this patch into separate commits which perform the following two steps independently:

* first commit: move the files from `bin` to `src/catkin_pkg/cli` without any modifications, keeping the individual files separately (not mixing them into a single module), the prefix `catkin_` can be removed during the rename since it then part of the package: e.g. `git mv bin/catkin_create_pkg src/catkin_pkg/cli/create_pkg`

* second commit: apply any changes to make the moved files real entry points

dmerejkowsky:
Thanks for the feedback!

I'll rewrite the PR to make it more readable

dmerejkowsky:
Note: I initially put everything into one module because it seemed there were functions in the scripts that could be useful elsewhere, such as `prompt_continue` in `generate_changelog.py`, but you may prefer having something like `catking_pkg/cli/utils.py`.

Last commit is up to you: I prefer not having entry points as "scripts", but that's a matter of personal preference IMHO.

src/catkin_pkg/cli/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
```
dirk-thomas: An empty init file should not contain a license since there is nothing in that file to license.

setup.py
```diff
@@ -15,13 +15,15 @@
...
-        'bin/catkin_tag_changelog',
-        'bin/catkin_test_changelog'
-    ],
+    entry_points={
+        'console_scripts' : [
+            'catkin_create_pkg         = catkin_pkg.cli.create_pkg:main',
+            'catkin_find_pkg           = catkin_pkg.cli.find_pkg:main',
+            'catkin_generate_changelog = catkin_pkg.cli.generate_changelog:main',
+            'catkin_tag_changelog      = catkin_pkg.cli.tag_changelog:main',
+            'catkin_test_changelog     = catkin_pkg.cli.test_changelog:main',
```
dirk-thomas: Please avoid the alignment. If a new console script is being added the whole block might need to be reformatted. And if e.g. two PRs add different ones they will conflict for no reason.

dirk-thomas:
Beside the small comments it looks good to me.

We could provide convenience scripts which would be callable without invoking `setup.py develop` which are only available in source and are not being installed? Not sure if anyone needs them though...

dmerejkowsky:
> We could provide convenience scripts

Not sure this is a good idea. I believe it would only bring confusion. Besides, people already have to run `setup.py develop`  (either by directly or with `pip`), in order to install requirements listed in the
`setup.py` file ...

dmerejkowsky:
Up ?

dirk-thomas:
Thanks for commenting. I will go ahead with the merge and afterwards add the scripts. Imo there are use cases where having them is helpful. E.g. when you already have the package as well as it's dependencies installed and just want to use the current state from master (without the need to setup a venv or install the package globally).

dirk-thomas:
By changing the scripts to entry points the error message from https://github.com/ros-infrastructure/catkin_pkg/pull/144/files#diff-4743da4e1fb236c03eab4325d1012a60R122 won't be used anymore. Instead the user would see the full stack trace. I updated the patch to maintain the previous error message.

Cherry-picked:

* moved files: a65d4d27b0dcc9d21aa769be612785f66c8f288e
* updated files, removed the `__main__` logic, and maintain the error message: f7dcb41c4c91bc18a29931eb4633ddd71fb58a0a

Added convenient scripts:

* d52bd668a1c0d78ee2d047f77176b135834d4be0

PR #145, opened 2016-05-11
allow for "revision" in package version and changelog

defunctzombie:
The revision is appended to the upstream version by a vendor to indicate
that the package is altered from upstream packaging.

This allows for users to create packages using the proper upstream
version with their own patch-set applied on top.

### DIFF
--- a/src/catkin_pkg/changelog.py
+++ b/src/catkin_pkg/changelog.py
@@ -295,7 +295,7 @@ def version_and_date_from_title(title):
     :returns: ``(str, datetime.datetime)``
     :raises: ``InvalidSectionTitle`` for non REP-0132 section titles
     '''
-    match = re.search(r'^([0-9]+\.[0-9]+\.[0-9]+)[ ]\((.+)\)$', title)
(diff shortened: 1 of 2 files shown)


defunctzombie:
This allows for `-[0-9]+[a-z]+` to optionally follow the version. The purpose for this is to make it possible for downstream consumers to patch and release modified versions without having to alter/increment the upstream version number or the package name/release or any other tricks. This is a typical approach and naming scheme seen on debian packages when altering from the base.

Happy to also relax the check here and allow for more flexibility in what goes after the `-`; this was just the first pass at bringing this feature in.

defunctzombie:
another possible regex is to allow `-[a-z]+[0-9]+` instead

dirk-thomas:
How does this affect all the other tools in e.g. the release toolchain which e.g. add a deb inc after the version, rosdistro expecting a certain version format, etc.? I would expect some of them to bail on this.

dirk-thomas:
I will close this for now due to no response. Please feel free to comment on the issue and it can be reopened.

PR #146, opened 2016-05-26
Converted python package declaration to ament style.

ernestmc:
Needed changes so the python package installs correctly under ament.


dirk-thomas:
I don't think it is necessary to move all the files in order to work with `ament`. Can you please describe why you think it is not possible without.

ernestmc:
@dirk-thomas if the python package ```catkin_pkg``` is inside the ```src``` folder ament isn't able to find it. I checked [ament_package](https://github.com/ament/ament_package) and followed a similar structure and so it worked.

dirk-thomas:
Ament uses plain setup.py files so I can't imagine why it wouldn't support that use case. If there is a corner case which is not covered right now that should be addressed instead of forcing packages to restructure.

ernestmc:
I agree. I'm the only one having this issue. I'll close this pull request and investigate further why is this happening.

PR #152, opened 2016-09-13
Install catkin_pkg.cli modules.

de-vri-es:
The new `catkin_pkg.cli` module was not installed by `setup.py`. This PR rectifies that.


dirk-thomas:
Thank you for catching this. For the record the new module was introduced in #144.

PR #154, opened 2017-01-04
Help ROS Users Switch to C++11

davetcoleman:
Help users such as [these guys](https://github.com/ros-industrial/universal_robot/pull/254) use MoveIt! in Kinetic that requires C++11. Switching is of course optional.


dirk-thomas:
Thanks.

PR #155, opened 2017-01-27
Added --python-modules argument to catkin_create_pkg to automatically…

peci1:
… create the setup.py file.

I'm open for discussion about this PR - I don't tell I've chosen the only correct way to cope with that :)

I can't make my mind if it'd be good to augment the `Package` class with slot `python_modules` which would actually read the list of modules from `setup.py`. I'm not sure if it wouldn't be too big change (and maybe too useless :) ).


dirk-thomas:
Maybe keeping it more simple would be enough to generate a skeleton package. `--python-module` could be a boolean option and if being passed only generate a single module with the name of the package. Thr setup file could even use [find_package](http://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages) to avoid the need to update the list manually (I haven't tried that for a ROS package yet).

peci1:
Well, the simple use-case was what I had on my mind. You can right now use it just as a boolean flag and it works.

The find_packages option is interesting, but maybe it's for a wider discussion, since it is something different than what's used in the docs...

dirk-thomas:
I will close this due to the long time of inactivity.

PR #156, opened 2017-01-30
add conflict with debian package `catkin`

wjwwood:
This is done to avoid the state where someone has our Python packages (the packages.ros.org `.deb` files) for `catkin_pkg` and/or `rospkg`, but are using the upstream "Debian Science" packages for ROS's C++ components (these are available from Ubuntu without any added apt sources). This situation leads to subtle bugs and should be avoided.

See:

- https://github.com/ros-infrastructure/rospkg/pull/102#issuecomment-273509972
- http://wiki.ros.org/UpstreamPackages

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -1,7 +1,7 @@
 [DEFAULT]
 Depends: python-argparse, python-dateutil, python-docutils
 Depends3: python3-dateutil, python3-docutils
-Conflicts: python3-catkin-pkg
-Conflicts3: python-catkin-pkg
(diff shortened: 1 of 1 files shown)


wjwwood:
@dirk-thomas do you want to wait to merge and deploy this when you do the modules break up? Otherwise, I'll merge this and push a release today.

dirk-thomas:
I would say merge but don't release. Will do a release after the split.

wjwwood:
Sounds good to me.

Issue #125, opened 2015-07-27
python-catkin-pkg and python3-catkin-pkg cannot both be installed side-by-side


jack-oquin:
The root problem seems to be the fact that both packages install scripts in `/usr/bin` with the same names, like `catkin_create_package` and `catkin_tag_changelog`.

This is clearly contrary to the Debian and Ubuntu intention to support both Python dialects side by side. I did some poking around to see how they recommend doing it. I didn't find any good documentation on that, so I experimented with the .deb releases of `python-sphinx` and `python3-sphinx`.

Those packages _can_ be installed side-by-side. Each version installs its scripts in different directories:
- /usr/share/sphinx/scripts/python
- /usr/share/sphinx/scripts/python3

Then, they symlink one or the other set of scripts into `/usr/bin`. So far, I don't see how that was done.

On my Trusty system, the `python` versions are selected. I don't know if that is because Python 2 is the default version, or because I installed `python-sphinx` first. I don't see anything in `/etc/alternatives` or `/var/lib/dpkg/alternatives`.

Surely, there is a reasonable solution to this packaging problem.

dirk-thomas:
This is indeed the common way to package Debian packages.

We use https://github.com/astraw/stdeb to package ROS Python packages and https://github.com/ros-infrastructure/ros_release_python to automate the process. Afaik stdeb does not support this at the moment. So we would either need to come up with a completely different approach to create the Debian packages or extend stdeb with that feature.

It is not likely that we will work on either of these two options in the near future. If anybody known an easier way to achieve this or want to work on this feature we would be very happy to support that effort.

jack-oquin:
It looks like the recommended Debian approach uses the ``debian/links`` mechanism:

* https://wiki.debian.org/Python/Packaging#Example_2:_Python_application

de-vri-es:
One approach I used for a similar problem when packaging software for Arch Linux was to split the package in a library and a scripts/tools package. Then the library packages can be installed side by side, and the scripts/tools package conflict and provide the same target (might be the name of one of the two packages).

This method only allows one of the scripts/tools package to be installed, but they should behave identical so that should not be a problem.

You would have the following packages:
* `python-catkin-pkg` (library)
* `python2-catkin-pkg` (library)
* `python-catkin-pkg-scripts` (conflicts with `python-catkin-pkg-scripts`)
* `python2-catkin-pkg-scripts` (conflicts with and provides `python-catkin-pkg-scripts`)

[...]

PR #157, opened 2017-02-03
split into two Debian packages

dirk-thomas:
Alternative approach to achieve #125.

For PIP a single package is being generated as before. Two separate Debian packages are generated - one only containing the Python modules (`python(3)-catkin-pkg-modules`), one containing only the scripts (`python(3)-catkin-pkg`) which depends on the first one.

Requires ros-infrastructure/ros_release_python#15.

@ros-infrastructure/ros_team Please review and comment.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,44 +1,53 @@
 #!/usr/bin/env python
 
-from setuptools import setup
-
 import os
 import sys
+
+from setuptools import setup
+
 source = os.path.join(os.path.dirname(__file__), 'src')
 sys.path.insert(0, source)
 
-from catkin_pkg import __version__
+from catkin_pkg import __version__  # noqa
 
-setup(
-    name='catkin_pkg',
-    version=__version__,
-    packages=['catkin_pkg', 'catkin_pkg.cli'],
-    package_dir={'': 'src'},
-    package_data={'catkin_pkg': ['templates/*.in']},
-    entry_points={
-        'console_scripts' : [
+kwargs = {
+    'name': 'catkin_pkg',
+    'version': __version__,
+    'packages': ['catkin_pkg', 'catkin_pkg.cli'],
+    'package_dir': {'': 'src'},
+    'package_data': {'catkin_pkg': ['templates/*.in']},
+    'entry_points': {
+        'console_scripts': [
             'catkin_create_pkg = catkin_pkg.cli.create_pkg:main',
             'catkin_find_pkg = catkin_pkg.cli.find_pkg:main',
             'catkin_generate_changelog = catkin_pkg.cli.generate_changelog:main_catching_runtime_error',
             'catkin_tag_changelog = catkin_pkg.cli.tag_changelog:main',
             'catkin_test_changelog = catkin_pkg.cli.test_changelog:main',
-        ]
-    },
-    author='Dirk Thomas',
-    author_email='dthomas@osrfoundation.org',
-    url='http://wiki.ros.org/catkin_pkg',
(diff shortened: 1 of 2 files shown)


cottsay:
@jpgr87 - we should consider how this might look in Fedora.

jpgr87:
According to the [guidelines](https://fedoraproject.org/wiki/Packaging:Python#Executables_in_.2Fusr.2Fbin), I think we just need to ship the scripts that go in `/usr/bin` in the python3 package and leave them out of the python2 package if the scripts provide identical functionality.

If they do both need to be installed, there's instructions for how to version the binaries to avoid collision.

In either case, we don't need the package to be split at all.

Debian [suggests](https://wiki.debian.org/Python/LibraryStyleGuide#Executables_and_library_packages) resolving this by either just putting the scripts in one package (similar to Fedora,) or providing a "caktin_pkg" deb package that ships the scripts, and having it depend on one of the python{2,3}-catkin_pkg library modules.  Doing something like that could remove the collision issue without having to split the source package, but you would presumably have to decide on python2 or python3 to execute the scripts.

dirk-thomas:
@jpgr87 Your comment seems to suggest almost exactly what we want to do: separate the modules into `python-catkin-pkg-modules` / `python3-catkin-pkg-modules` and having a package for the scripts which depend on it (``python-catkin-pkg`). We only keep two packages - one with the Python 2 scripts and another one with the Python 3 scripts in order to not break existing users which have exactly that package installed. If we wouldn't need to consider that than we could indeed drop on of the script packages since the users don't care which Python version it uses in the shebang lines.

dirk-thomas:
When calling `apt update && apt dist-upgrade` the new dependency fails to be installed (since it provides the same files as the existing `python-catkin-pkg` package:

```
Unpacking python-catkin-pkg-modules (0.3.0-1) ...
dpkg: error processing archive /var/cache/apt/archives/python-catkin-pkg-modules_0.3.0-1_all.deb (--unpack):
trying to overwrite '/usr/lib/python2.7/dist-packages/catkin_pkg/metapackage.py', which is also in package python-catkin-pkg 0.2.10-2
Preparing to unpack .../python-catkin-pkg_0.3.0-1_all.deb ...
Unpacking python-catkin-pkg (0.3.0-1) over (0.2.10-2) ...
Processing triggers for man-db (2.7.5-1) ...
Errors were encountered while processing:
/var/cache/apt/archives/python-catkin-pkg-modules_0.3.0-1_all.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

The user than has to forcefully install the new package version / the dependency. Does anyone have an idea how we can make this more convenient? Otherwise I would create a question on answers.ros.org with problem and provide an answer for it. But I expect many users will have a problem with this...

Btw. the new packages are already available in the public repo on packages.ros.org so it is only a question of time when the first questions will come up.

de-vri-es:
> The user than has to forcefully install the new package version / the dependency. Does anyone have an idea how we can make this more convenient?

pacman packages can have versioned conflict and replace keys:
```
conflicts=('python-catkin-pkg<0.3')
replaces=('python-catkin-pkg<0.3')
```

Then the package manager will inform the user of the conflict explicitly and with the replace key it will ask the user to replace the old package. I'm not sure if there is an equivalent for Debian though.

dirk-thomas:
@de-vri-es Thank you for the hint - in Debian it is exactly the same (minor syntax differences). I remember we added support for those in the ROS manifest but never had an actual use for it. I guess I understand now better what they are intended for. I just released a new version 0.3.1 which uses those and it looks like it is upgrading smoothly now.

PR #159, opened 2017-02-06
improve exception message

### LINKED ISSUE #158: [catkin_generate_changelog] Reason of the error unclear 'Could neither find section "None" nor any other section'
You should be able to reproduce the issue by:

```
git clone git@github.com:ros-planning/moveit.git
cd moveit
git checkout jade-devel
git checkout fa3812c630920e29a7208167320f10eaac6b1447
catkin_generate_changelog
Found packages: chomp_motion_planner, moveit, moveit_commander, moveit_controller_manager_example, moveit_core, moveit_experimental, moveit_fake_controller_manager, moveit_kinematics, moveit_planners, moveit_planners_chomp, moveit_planners_ompl, moveit_plugins, moveit_ros, moveit_ros_benchmarks, moveit_ros_benchmarks_gui, moveit_ros_control_interface, moveit_ros_manipulation, moveit_ros_move_group, moveit_ros_perception, moveit_ros_planning, moveit_ros_planning_interface, moveit_ros_robot_interaction, moveit_ros_visualization, moveit_ros_warehouse, moveit_runtime, moveit_setup_assistant, moveit_simple_controller_manager
Querying commit information since latest tag...
Updating forthcoming section of changelog files...
- updating 'moveit_plugins/moveit_ros_control_interface/CHANGELOG.rst'
ERROR: Could neither find section "None" nor any other section
```

I looked at [CHANGELOG.rst](https://github.com/ros-planning/moveit/blob/fa3812c630920e29a7208167320f10eaac6b1447/moveit_plugins/moveit_ros_control_interface/CHANGELOG.rst) in the error message but I can't tell what's wrong/missing there. I also looked at CHANGELOG.rst files in other packages [in the same folder [moveit_plugins](https://github.com/ros-planning/moveit/tree/fa3812c630920e29a7208167320f10eaac6b1447/moveit_plugins) but they seem good to me.

So what can be the issue here. Also we might want to improve error message.

dirk-thomas:
Improve exception message to mention the section name which was searched for.

Fixes #158.

@ros-infrastructure/ros_team Please review.

### DIFF
--- a/src/catkin_pkg/changelog_generator.py
+++ b/src/catkin_pkg/changelog_generator.py
@@ -176,7 +176,10 @@ def update_changelog_file(data, tag2log_entries, vcs_client=None, skip_contribut
                     data = data[:match.start()] + block + '\n' + data[match.start():]
                     break
             if not match:
-                raise RuntimeError('Could neither find section "%s" nor any other section' % tag.name)
+                if tag.name is None:
+                    raise RuntimeError('Could not find section "%s"' % next_tag.name)
+                else:
+                    raise RuntimeError('Could neither find section "%s" nor any other section' % tag.name)
         return data
 
 


130s:
Thanks for the quick action!

So, what's "section"? is the first question I had when I saw the error message, which I couldn't figure out until I took a look at the [inline comment](https://github.com/ros-infrastructure/catkin_pkg/blob/e0ade071e4651a4b18b2070a2eb33a538a1905ed/src/catkin_pkg/changelog_generator.py#L164). Can we print something like "version section"?

dirk-thomas:
For `reStructuredText` the term section is clearly defined. Imo the current message already provides enough information to identify the problem since it includes the version number. If you feel strongly about this please feel free to create a PR and we can see what others think.

130s:
I agree for the case where the cause was the missing version section, this PR gives much better output (see https://github.com/ros-infrastructure/catkin_pkg/issues/158#issuecomment-277852080).

I don't know about the case where the following error occurs. I'm afraid users will have no idea what "section" is.

> ERROR: Could neither find section "None" nor any other section

dirk-thomas:
> ERROR: Could neither find section "None" nor any other section

This error message can't happen anymore with the patch applied.

PR #160, opened 2017-02-27
remove trailing whitespace

wahtak:
This PR cleanes up a trainling whitespace in the package.xml template, as it shows up during code review of generated package.xml files.


dirk-thomas:
Thank you for the patch.

Issue #161, opened 2017-03-28
Drop package naming convention warnings for non-catkin packages.


mikepurvis:
Packages which have been third-party-released into the ROS ecosystem shouldn't be subject to as stringent naming conventions. Would a patch be welcome which allows dashes when the build type is non-catkin?

dirk-thomas:
The current restriction ensures that package names are unique and can be released without a collision. If `foo_bar` as well as `foo-bar` are valid package names they will collide when being released as Debian packages.

Since third-party packages don't have a manifest file upstream they can easily use the underscore variation in the manifest. So it is not preventing them from being packaged at the moment.

But I agree to your point that we should not enforce conventions on packages if not necessary. Therefore I am not sure if the pro outweighs the con. Let see what other think: @ros-infrastructure/ros_team

mikepurvis:
The situation for us is that we have internal packages which have `<depend>foo-bar</depend>` in their manifest, and that has traditionally been a package that we've released (or imported) as a standalone deb. However, now for various reasons we are "third party releasing" some of these packages and building them as part of our distribution bundle, dropping the dependency on the standalone deb package.

This applies particularly to Python packages, in conjunction with [catkin_tools_python](https://github.com/mikepurvis/catkin_tools_python). It feels awkward and wrong to visit a consuming package and change a dependency from `python-bar-baz` to `python_bar_baz` just to comply with this naming requirement. The canonical name of such a dep has long been established as the "debian" name with dashes, and the consuming package shouldn't care whether it's ultimately being provided by a system package or a catkin workspace package.

mikepurvis:
It's a fairly trivial change— I've opened PR #162, so let's take the discussion there.

PR #162, opened 2017-03-28
Allow dashes in package names for non-catkin packages.

mikepurvis:
Rationale given in #161.

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -177,8 +177,17 @@ class Package(object):
         # accepting upper case letters and hyphens only for backward compatibility
         if not re.match('^[a-zA-Z0-9][a-zA-Z0-9_-]*$', self.name):
             errors.append('Package name "%s" does not follow naming conventions' % self.name)
-        elif not re.match('^[a-z][a-z0-9_]*$', self.name):
-            new_warnings.append('Package name "%s" does not follow the naming conventions. It should start with a lower case letter and only contain lower case letters, digits and underscores.' % self.name)
+        else:
+            if self.has_buildtool_depend_on_catkin():
+                if not re.match('^[a-z][a-z0-9_]*$', self.name):
+                    new_warnings.append(
+                            'Catkin package name "%s" does not follow the naming conventions. It should start with '
+                            'a lower case letter and only contain lower case letters, digits and underscores.' % self.name)
+            else:
+                if not re.match('^[a-z][a-z0-9_-]*$', self.name):
+                    new_warnings.append(
+                            'Non-catkin package name "%s" does not follow the naming conventions. It should start with'
+                            'a lower case letter and only contain lower case letters, digits, underscores, and dashes.' % self.name)
 
         if not self.version:
             errors.append('Package version must not be empty')
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -137,6 +137,18 @@ class Package(object):
         """
         return 'catkin' in [d.name for d in self.buildtool_depends]
 
+    def build_type(self):
+        """
+        Returns value of export/build_type element, or 'catkin' if unspecified.
+
+        :returns: package build type
+        :rtype: str
+        """
+        for e in self.exports:
+            if e.tagname == 'build_type':
+                return e.content
+        return 'catkin'
+
     def has_invalid_metapackage_dependencies(self):
         """
         Returns True if this package has invalid dependencies for a metapackage
@@ -178,7 +190,7 @@ class Package(object):
         if not re.match('^[a-zA-Z0-9][a-zA-Z0-9_-]*$', self.name):
             errors.append('Package name "%s" does not follow naming conventions' % self.name)
         else:
-            if self.has_buildtool_depend_on_catkin():
+            if self.build_type() == 'catkin':
                 if not re.match('^[a-z][a-z0-9_]*$', self.name):
                     new_warnings.append(
                             'Catkin package name "%s" does not follow the naming conventions. It should start with '
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -196,13 +196,13 @@ class Package(object):
             if self.get_build_type() == 'catkin':
                 if not re.match('^[a-z][a-z0-9_]*$', self.name):
                     new_warnings.append(
-                            'Catkin package name "%s" does not follow the naming conventions. It should start with '
-                            'a lower case letter and only contain lower case letters, digits, and underscores.' % self.name)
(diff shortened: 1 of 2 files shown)


test/test_package.py
```diff
@@ -164,6 +185,7 @@ def test_validate_package(self):
         self.assertRaises(InvalidPackage, Package.validate, pack)
         pack.authors = []
         pack.validate()
+
```
mikepurvis: I know we don't normally like random whitespace changes, but this is within the same function, and makes the test more consistent and readable.

mikepurvis:
Via #161, Dirk's concern:

> The current restriction ensures that package names are unique and can be released without a collision. If foo_bar as well as foo-bar are valid package names they will collide when being released as Debian packages.

Our situation:

> ... we have internal packages which have <depend>foo-bar</depend> in their manifest, and that has traditionally been a package that we've released (or imported) as a standalone deb. However, now for various reasons we are "third party releasing" some of these packages and building them as part of our distribution bundle, dropping the dependency on the standalone deb package.
>
> This applies particularly to Python packages, in conjunction with catkin_tools_python. It feels awkward and wrong to visit a consuming package and change a dependency from python-bar-baz to python_bar_baz just to comply with this naming requirement. The canonical name of such a dep has long been established as the "debian" name with dashes, and the consuming package shouldn't care whether it's ultimately being provided by a system package or a catkin workspace package.

@ros-infrastructure/ros_team Thoughts on this change?

src/catkin_pkg/package.py
```diff
@@ -177,8 +177,17 @@ def validate(self, warnings=None):
         # accepting upper case letters and hyphens only for backward compatibility
         if not re.match('^[a-zA-Z0-9][a-zA-Z0-9_-]*$', self.name):
             errors.append('Package name "%s" does not follow naming conventions' % self.name)
-        elif not re.match('^[a-z][a-z0-9_]*$', self.name):
-            new_warnings.append('Package name "%s" does not follow the naming conventions. It should start with a lower case letter and only contain lower case letters, digits and underscores.' % self.name)
+        else:
+            if self.has_buildtool_depend_on_catkin():
```
wjwwood: What's the rationale on using this check versus the `build_type` of the package (no `build_type` implies `catkin`)?

mikepurvis: Oh, mostly just that there was the existing function to check for it, so it was easy to implement. If you'd prefer to check for `<export><build_type></build_type></export>`, it can be switched to that (I would add a similar helper function).

wjwwood: I don't have a strong feeling one way or the other, but `build_type` seems purpose built for this, but I thought maybe I wasn't considering some corner case which asks for this style check instead.

mikepurvis: Nope, purely convenience.

dirk-thomas:
Please also create a PR to amend [REP 127](http://www.ros.org/reps/rep-0127.html#name). I am not sure how we should deal with it though. Technically it would need to be a new format since any implementation of the currently specified format 2 would be not be accurate anymore with this change. On the other hand we don't want to create a new format just for this...

mikepurvis:
The current text simply gives the strictest possible expectations with no mention of warnings, or the exceptions for legacy support already in `catkin_pkg`:

```
The package name must start with a letter and contain only lowercase alphabetic, numeric or underscore characters [1]. The package name should be unique within the ROS community. It may differ from the folder name into which it is checked out, but that is not recommended.
```

So I can add a paragraph expanding on this, but I don't really agree that the underlying format is changing (particularly since the *recommended* character set the same, and it's simply an exception carved out for what was previously a warning).

dirk-thomas:
All I am saying is that if anybody implemented a parser based on the spec a manifest file with your relaxed name might be rejected. E.g. the current schema will not pass for your relaxed manifests and will need to be updated: http://download.ros.org/schema/package_format2.xsd

I am not advocating to create a format 3 for this. I just want to clarify that we are bending the rules here. Therefore it might be good to announce this on the mailing list to allow others to provide feedback on this.

mikepurvis:
Absolutely, makes sense. Interesting point about the XSD validator— that's already stricter than the validation which occurs in `catkin_pkg`.

dirk-thomas:
> Absolutely, makes sense. Interesting point about the XSD validator— that's already stricter than the validation which occurs in `catkin_pkg`.

Yeah, the schema tries to model the spec as exactly as possible.

mikepurvis:
PR up, see https://github.com/ros-infrastructure/rep/pull/126

src/catkin_pkg/package.py
```diff
@@ -137,6 +137,18 @@ def has_buildtool_depend_on_catkin(self):
...
+        """
+        Returns value of export/build_type element, or 'catkin' if unspecified.
+
+        :returns: package build type
+        :rtype: str
+        """
+        for e in self.exports:
+            if e.tagname == 'build_type':
+                return e.content
+        return 'catkin'
```
mikepurvis: @dirk-thomas @wjwwood PR updated with a `build_type` check instead of buildtool. Could make this a property rather than a method by adding it to `getattr`, up to you guys if you'd prefer it there.

wjwwood: I think this is fine. If you're worried about it getting confused with the properties you could just call it `get_build_type()`. I'm fine with either.

dirk-thomas: If the tag appears multiple times I think the function should do something about it - maybe raise?

mikepurvis: Changed to `get_build_type`, added raise on multiple instances of the element. That's an increase in strictness from previously, although this is in the more freeform `<export>` section, and REP 127 doesn't actually have anything to say about `<build_type>`.

Maybe as part of https://github.com/ros-infrastructure/rep/pull/126 it should?

mikepurvis:
I think this is good to go in, unless we want to hold it and merge together with https://github.com/ros-infrastructure/rep/pull/126.

dirk-thomas:
Waiting for a decision on [this](https://github.com/ros-infrastructure/rep/pull/126#issuecomment-293634573) question.

mikepurvis:
Okay, well, I've committed all the time I can to this. We'll monkey patch it on our side in the meantime and you guys can merge or not when you're ready.

mikepurvis:
For interested parties, here's a simple monkey-patch to silence package validation errors:

```
# Patch over Package.validate with a version that silences warnings, pending:
# https://github.com/ros-infrastructure/catkin_pkg/pull/162
old_validate = catkin_pkg.package.Package.validate
catkin_pkg.package.Package.validate = lambda self, warnings: old_validate(self, [])
```

PR #164, opened 2017-04-10
Make C++11 suggestion use set_directory_properties

mikepurvis:
Follow-on from #154. FYI @davetcoleman

### DIFF
--- a/src/catkin_pkg/templates/CMakeLists.txt.in
+++ b/src/catkin_pkg/templates/CMakeLists.txt.in
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(@name)
 
 ## Compile as C++11, supported in ROS Kinetic and newer
-# set_directory_properties(PROPERTIES COMPILE_OPTIONS "-std=c++11")
+# add_compile_options(-std=c++11)
 
 ## Find catkin macros and libraries
 ## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
--- a/src/catkin_pkg/templates/CMakeLists.txt.in
+++ b/src/catkin_pkg/templates/CMakeLists.txt.in
@@ -1,8 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(@name)
 
-## Add support for C++11, supported in ROS Kinetic and newer
-# add_definitions(-std=c++11)
+## Compile as C++11, supported in ROS Kinetic and newer
+# set_directory_properties(PROPERTIES COMPILE_OPTIONS "-std=c++11")
 
 ## Find catkin macros and libraries
 ## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)


davetcoleman:
I'm not sure where the long conversation was anymore, but within the moveit community we decided the current approach is better. Why do you think this change is necessary?

@v4hn @de-vri-es

v4hn:
On Mon, Apr 10, 2017 at 09:24:53PM -0700, Dave Coleman wrote:
> I'm not sure where the long conversation was anymore,

I think you mean https://github.com/ros-planning/moveit_planners/pull/73

> but within the moveit community we decided the current approach is better.

No, we didn't. We never really talked about `add_definitions` because this abuses the `DEFINITIONS` mechanism of cmake and is not a good idea.
We use `add_compile_options(-std=c++11)` with MoveIt. This requires cmake 2.8.12 - the current template only requires 2.8.3, but I don't think this is a problem anymore.

`add_compile_options()` and `set_directory_properties(PROPERTIES COMPILE_OPTIONS "")` are pretty much identical. As far as I can see they where even introduced in the same cmake version.
I clearly prefer the former because it's easier to read, but this is a matter of preferences.

de-vri-es:
> `add_compile_options()` and `set_directory_properties(PROPERTIES COMPILE_OPTIONS "")` are pretty much identical. As far as I can see they where even introduced in the same cmake version. I clearly prefer the former because it's easier to read, but this is a matter of preferences.

Indeed, judging from the documentation `add_compile_options` is simply a shorthand for setting the relevant directory property. To me  `add_compile_options` looks much prettier and more to the point.

I pretty much completely agree with @v4hn

mikepurvis:
`add_definitions` has no scope, so it leaks into other packages when building with `catkin_make`.

I'm ambivalent between `add_compile_options` and `set_directory_properties`; the suggestion here simply follows what is done in ros_comm: https://github.com/ros/ros_comm/search?q=set_directory_properties

v4hn:
I believe the only crucial difference between the latter two is that `add_compile_options` **appends** the flag, while `set_directory_properties` overwrites them.

I believe appending is the better choice for this package template.

dirk-thomas:
`add_compile_options` sounds like the better choice than. Especially for new users which are likely going to just use what the template provides.

mikepurvis:
Changed.

Next question— when do we make this on by default?

dirk-thomas:
Setting the compile options will only affect the compiler flags of this package. If the exposed headers also require a C++11 compiler the downstream packages need to specify that too. I think it would be good to mention that in the comment to not result in false expectations.

> Next question— when do we make this on by default?

That will be up for an REP (likely [REP 3](http://www.ros.org/reps/rep-0003.html)) to decide.

davetcoleman:
Wow I thought I had mimicked what we were doing in MoveIt!, but I must have referenced the wrong CMakeLists at the time. I stand corrected!

dirk-thomas:
Thank you for the patch!

PR #165, opened 2017-04-11
Check python version in a 2.6 and below compatible manner.

de-vri-es:
Changed the way python2/3 is detected. Now uses the same check as `changelog.py` [1].

This should be compatible with python versions below 2.7, which the old check was not (see also [2]). 

[1] https://github.com/ros-infrastructure/catkin_pkg/blob/fe86f9277eef68a0ebe0c16047e795f3b1533f54/src/catkin_pkg/changelog.py#L46
[2] https://github.com/ros-infrastructure/catkin_pkg/commit/84b14a0668a9728525d2bfe7eb33abd915e555e5#commitcomment-21710999

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -371,7 +371,7 @@ def parse_package(path, warnings=None):
     # Force utf8 encoding for python3.
     # This way unicode files can still be processed on non-unicode locales.
     kwargs = {}
-    if sys.version_info.major >= 3:
+    if sys.version_info[0] >= 3:
         kwargs['encoding'] = 'utf8'
 
     with open(filename, 'r', **kwargs) as f:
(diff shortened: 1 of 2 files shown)


src/catkin_pkg/package.py
```diff
@@ -371,7 +373,7 @@ def parse_package(path, warnings=None):
     # Force utf8 encoding for python3.
     # This way unicode files can still be processed on non-unicode locales.
     kwargs = {}
-    if sys.version_info.major >= 3:
+    if _py3:
```
dirk-thomas: Since this is only being used once please move the condition to here instead of using a variable.

Since we only care about the major version can you please update it to only check for `[0] >= 3`.

de-vri-es: Fair enough, though I figured keeping it in line with `changelog.py` made sense. Instead I changed the check in `changelog.py` now.

dirk-thomas:
Thank you!

Issue #167, opened 2017-05-09
add include_directories() to template regardless of whether or not depend specified


jwhendy:
Greetings,

**tl;dr:** I'd like to suggest adding the following to `CMakeLists.txt` when `catkin_create_pkg` is called, regardless of whether or not the user specifies dependencies (perhaps as a commented out section?).
```
include_directories(
  ${catkin_INCLUDE_DIRS}
)
```
---

I'm going through [*A Gentle Introduction to ROS*](https://cse.sc.edu/~jokane/agitr/) and was hung up for quite some time on why the first `Hello, world` exercise wouldn't build, failing to find `ros/ros.h` from my include statement. Eventually I stumbled on [the solution](http://answers.ros.org/question/237494/fatal-error-rosrosh-no-such-file-or-directory/), which was to add this line to `CMakeLists.txt`:

```
include_directories(
  ${catkin_INCLUDE_DIRS}
)
```

I've also been through the ROS beginner tutorials in which one also creates a package, and never ran into this issue. I was determined to figure out why, and after much hunting, the only difference is whether or not `catkin_create_pkg` is called with a depend or not.

[ROS wiki](http://wiki.ros.org/ROS/Tutorials/CreatingPackage#ROS.2BAC8-Tutorials.2BAC8-catkin.2BAC8-CreatingPackage.Creating_a_catkin_Package)
```
$ catkin_create_pkg beginner_tutorials std_msgs rospy roscpp
```

*A Gentle Introduction to ROS* (Section 3.1, pg 39)
```
$ catkin_create_pkg agitr
```

I think it would be awesome (if nothing else, for beginners!) to have the `include_directories()` section included just as other template-style (but commented) sections already exist in `CMakeLists.txt`. Here's my reproducible example:

```
$ cd ~/catkin_ws/src
$ catkin_create_pkg beginner_tutorials roscpp
$ catkin_create_pkg beginner_tutorials2

### name-only differences omitted
$ diff --suppress-common-lines beginner_tutorials/CMakeLists.txt beginner_tutorials2/CMakeLists.txt 
10,12c10
< find_package(catkin REQUIRED COMPONENTS
<   roscpp
< )
---
> find_package(catkin REQUIRED)
105,106c103,104
< #  LIBRARIES beginner_tutorials
< #  CATKIN_DEPENDS roscpp
---
> #  LIBRARIES beginner_tutorials2
> #  CATKIN_DEPENDS other_catkin_pkg
117,119d114
[...]

PR #168, opened 2017-05-09
updating template to populate include_directories() with #

jwhendy:
This takes a stab at #167 . I thought of two routes and ended up modeling this after the behavior for `catkin_package()` in the template; providing no dependencies to `catkin_create_pkg` will yield this:

```
## Specify additional locations of header files
## Your package locations should be listed before other locations
include_directories(
#  include
#  ${catkin_INCLUDE_DIRS}
)
```
So, you get an empty call this way.

---

Alternatively, the current behavior is to initialize `result` with this:
```
# include_directories(include)
```
Any actual includes are wrapped in `include_directories()` and appended *if* they exist. So, my second route would have been:
```
def _create_include_macro(package_template):
    result = '# include_directories(include)'
    includes = []
    if package_template.catkin_deps:
        includes.append('${catkin_INCLUDE_DIRS}')
    else:
        includes.append('#  ${catkin_INCLUDE_DIRS}')
...
```
I thought the first was cleaner, but the latter is more true to the original code.

### DIFF
--- a/src/catkin_pkg/package_templates.py
+++ b/src/catkin_pkg/package_templates.py
@@ -294,10 +294,12 @@ def _create_targetlib_args(package_template):
 
 
 def _create_include_macro(package_template):
-    result = '# include_directories(include)'
+    result = '#  include'
     includes = []
     if package_template.catkin_deps:
         includes.append('${catkin_INCLUDE_DIRS}')
+    else:
+        includes.append('#  ${catkin_INCLUDE_DIRS}')
     if package_template.boost_comps:
         includes.append('${Boost_INCLUDE_DIRS}')
     if package_template.system_deps:
@@ -309,7 +311,7 @@ def _create_include_macro(package_template):
         if deplist:
             result += '\n# TODO: Check names of system library include directories (%s)' % ', '.join(deplist)
     if includes:
-        result += '\ninclude_directories(\n  %s\n)' % '\n  '.join(includes)
+        result += '\n%s' % '\n  '.join(includes)
     return result
 
 
--- a/src/catkin_pkg/package_templates.py
+++ b/src/catkin_pkg/package_templates.py
@@ -294,7 +294,6 @@ def _create_targetlib_args(package_template):
 
 
 def _create_include_macro(package_template):
-    result = ''
     includes = ['# include']
     includes.append(('  ' if package_template.catkin_deps else '# ') + '${catkin_INCLUDE_DIRS}')
     if package_template.boost_comps:
@@ -308,6 +307,7 @@ def _create_include_macro(package_template):
             todo_incl = '# TODO: Check names of system library include directories'
             includes.append(todo_incl + (' (%s)' % ', '.join(deplist)))
             includes.extend(['  ${%s_INCLUDE_DIRS}' % sysdep for sysdep in deplist])
+    result = ''
     if includes:
         result += '%s' % '\n'.join(includes)
     return result
--- a/src/catkin_pkg/package_templates.py
+++ b/src/catkin_pkg/package_templates.py
@@ -309,7 +309,7 @@ def _create_include_macro(package_template):
             includes.extend(['  ${%s_INCLUDE_DIRS}' % sysdep for sysdep in deplist])
     result = ''
     if includes:
-        result += '%s' % '\n'.join(includes)
+        result += '\n'.join(includes)
     return result
 
 def _create_depend_tag(dep_type,
--- a/src/catkin_pkg/package_templates.py
+++ b/src/catkin_pkg/package_templates.py
@@ -294,26 +294,22 @@ def _create_targetlib_args(package_template):
 
 
 def _create_include_macro(package_template):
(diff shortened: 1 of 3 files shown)


src/catkin_pkg/package_templates.py
```diff
@@ -294,10 +294,12 @@ def _create_targetlib_args(package_template):
 
 
 def _create_include_macro(package_template):
-    result = '# include_directories(include)'
+    result = '#  include'
```
dirk-thomas: Should this go into `includes` instead so that it gets indented the same way? Like this: `includes = ['# include']`

Nitpick: can you please remove one of the two spaces. The `#` should be followed by a single space.

src/catkin_pkg/package_templates.py
```diff
@@ -294,10 +294,12 @@ def _create_targetlib_args(package_template):
...
     includes = []
     if package_template.catkin_deps:
         includes.append('${catkin_INCLUDE_DIRS}')
+    else:
+        includes.append('#  ${catkin_INCLUDE_DIRS}')
```
dirk-thomas: This certainly works. In general I try to avoid repetition whenever possible. Can you update this block to something like this:

```
include = '${catkin_INCLUDE_DIRS}'
if not package_template.catkin_deps:
    include = '# ' + include
includes.append(include)
```

src/catkin_pkg/package_templates.py
```diff
@@ -309,7 +311,7 @@ def _create_include_macro(package_template):
         if deplist:
             result += '\n# TODO: Check names of system library include directories (%s)' % ', '.join(deplist)
```
dirk-thomas: This message will now appear above all the includes and not anymore directly above `deplist`. Maybe just append this comment to `includes` (without the leading newline)?

src/catkin_pkg/package_templates.py
```diff
@@ -309,7 +311,7 @@ def _create_include_macro(package_template):
...
     if includes:
-        result += '\ninclude_directories(\n  %s\n)' % '\n  '.join(includes)
+        result += '\n%s' % '\n  '.join(includes)
```
dirk-thomas: The logic doesn't indent the first item of the list? Maybe if we get of `results` all together (see the other comments) the lines 313-315 can be folded into this:

```
return '\n'.join(['  %s' % include for include in includes])
```

src/catkin_pkg/package_templates.py
jwhendy: Yes, I wasn't sure how to accomplish that. I thought of just initializing `result` to an empty string and having everything be appended from `includes`?

Regarding the spaces, I was just following what looked like the convention. Here's pasted directly from the first instance of this sort in [CMakeLists.txt.in](https://raw.githubusercontent.com/ros-infrastructure/catkin_pkg/master/src/catkin_pkg/templates/CMakeLists.txt.in):
```
# add_message_files(
#   FILES
#   Message1.msg
#   Message2.msg
# )
```
Let me know what's considered proper style and I'll make it so!

src/catkin_pkg/package_templates.py
jwhendy: For sure. I should have mentioned, not only am I a complete ROS noob, but my `python` consists of 2 fully and 1 partially completed coursera class... I suppose free, but not-super-foo help, is still free though :)

src/catkin_pkg/package_templates.py
jwhendy: Good catch; I missed that completely.

src/catkin_pkg/package_templates.py
jwhendy: I'll have to brush up on (aka learn) my list parsing (`.append` and `join()`) it seems!

dirk-thomas:
Please update the tests as well to pass for the modified content.

jwhendy:
@dirk-thomas I'm not very familiar with this; I take it I'm to change entries like [line 118](https://github.com/ros-infrastructure/catkin_pkg/blob/master/test/test_templates.py#L118) in `test_templates.py` to contain the correct output of the new [`_create_inlcude_macro`](https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/package_templates.py#L296)?

dirk-thomas:
Yes, correct.

jwhendy:
Cool. I figured that was the case and started about it. Thanks for confirming, and with my recent push it looks like things are well!

Want to take a look at [this line](https://github.com/jwhendy/catkin_pkg/blob/include_dirs_fix/src/catkin_pkg/package_templates.py#L3080)? I noted that the `# TODO...` line was pretty long and opted to put the note in a variable and cat together to make it shorter. Not sure if spurious variables are frowned upon...

dirk-thomas:
Thanks for the patch and for iterating on it.

I applied some minor changes in a8bbac218506032f5c2d8236e908d187d1180c51 and 7776abac193ca9216cebeb65d16252b03280f497.

Issue #169, opened 2017-05-10
Parallel find_packages_allowing_duplicates?


mikepurvis:
For large workspaces (600+ packages), the serialized parsing of `package.xml` files in `find_packages_allowing_duplicates` can start to be non-trivial:

https://github.com/ros-infrastructure/catkin_pkg/blob/a4cb118d2da9a218f5e5b2216f5b1fdb93c0c5c5/src/catkin_pkg/packages.py#L109-L110

Without changing the interface to the function, would we consider allowing this work to be spread over multiple threads or processes, possibly triggered by some threshold in number of packages?

dirk-thomas:
Absolutely! The caller shouldn't care how the requested information is being gathered. If that loop can be parallelized that would be great.

mikepurvis:
A naive threading implementation is slower than the simple loop, so it's not IO bound. I get ~1.5s with the current implementation, and <0.5s running it with a multiprocessing map. I'll send a PR shortly.

mikepurvis:
The simplest implementation is like so:

```
package_paths = find_package_paths(basepath, exclude_paths=exclude_paths, exclude_subspaces=exclude_subspaces)
processed = multiprocessing.Pool(4).map(parse_package, package_paths)
return dict(zip(package_paths, parsed_packages))
```

However, to preserve the behaviour of the `warnings` attribute requires passing an extra attribute into the map for it and then manually aggregating the results, which necessitates some additional wrapping, unfortunately.

dirk-thomas:
Addressed by #171.

PR #170, opened 2017-05-10
Use multiprocessing when parsing packages.

mikepurvis:
As discussed in #169.

It's a nuisance having to put the wrapper/helper class up at module level, but multiprocessing can only pickle top level functions and objects, so I don't see an easy way around this. The class could potentially be avoided by defining separate helpers to always capture/not capture warnings, but that's a matter of taste.

### DIFF
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -102,7 +102,7 @@ class _PackageParser(object):
     def __call__(self, path):
         warnings = [] if self.capture_warnings else None
         parsed_package = parse_package(os.path.join(self.basepath, path), warnings=warnings)
-        return path, parsed_package, warnings
+        return (path, parsed_package), warnings
 
 
 def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subspaces=False, warnings=None):
@@ -117,11 +117,11 @@ def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subs
     :returns: A dict mapping relative paths to ``Package`` objects ``dict``
     """
     package_paths = find_package_paths(basepath, exclude_paths=exclude_paths, exclude_subspaces=exclude_subspaces)
-    parser = _PackageParser(basepath, isinstance(warnings, list))
-    results = multiprocessing.Pool(4).map(parser, package_paths)
+    parser = _PackageParser(basepath, warnings is not None)
+    path_parsed_packages, warnings_lists = zip(*multiprocessing.Pool().map(parser, package_paths))
     if parser.capture_warnings:
-        map(warnings.extend, [package_warnings for _, _, package_warnings in results])
-    return dict([(path, parsed_package) for path, parsed_package, _ in results])
+        map(warnings.extend, warnings_lists)
+    return dict(path_parsed_packages)
 
 
 def verify_equal_package_versions(packages):
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -34,6 +34,7 @@
 Library to find packages in the filesystem.
 """
 
+import multiprocessing
 import os
 from .package import parse_package, PACKAGE_MANIFEST_FILENAME
 
@@ -93,6 +94,17 @@ def find_packages(basepath, exclude_paths=None, exclude_subspaces=False, warning
     return packages
 
 
+class _PackageParser(object):
+    def __init__(self, basepath, capture_warnings):
+        self.basepath = basepath
+        self.capture_warnings = capture_warnings
+
+    def __call__(self, path):
+        warnings = [] if self.capture_warnings else None
+        parsed_package = parse_package(os.path.join(self.basepath, path), warnings=warnings)
+        return path, parsed_package, warnings
+
+
 def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subspaces=False, warnings=None):
     """
     Crawls the filesystem to find package manifest files and parses them.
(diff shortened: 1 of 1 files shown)


src/catkin_pkg/packages.py
```diff
@@ -104,11 +115,12 @@ def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subs
     :param warnings: Print warnings if None or return them in the given list
     :returns: A dict mapping relative paths to ``Package`` objects ``dict``
     """
-    packages = {}
     package_paths = find_package_paths(basepath, exclude_paths=exclude_paths, exclude_subspaces=exclude_subspaces)
-    for path in package_paths:
-        packages[path] = parse_package(os.path.join(basepath, path), warnings=warnings)
-    return packages
+    parser = _PackageParser(capture_warnings=isinstance(warnings, list))
```
dirk-thomas: Should this check we changed to:

`capture_warnings=warnings is not None`

in order to allow "duck typing"?

src/catkin_pkg/packages.py
```diff
@@ -104,11 +115,12 @@ def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subs
...
+    results = multiprocessing.Pool(4).map(parser, package_paths)
+    if parser.capture_warnings:
+        map(warnings.extend, [package_warnings for _, _, package_warnings in results])
```
dirk-thomas: Maybe use `results[2]` to avoid the underscores?

src/catkin_pkg/packages.py
```diff
@@ -104,11 +115,12 @@ def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subs
     :param warnings: Print warnings if None or return them in the given list
     :returns: A dict mapping relative paths to ``Package`` objects ``dict``
     """
-    packages = {}
     package_paths = find_package_paths(basepath, exclude_paths=exclude_paths, exclude_subspaces=exclude_subspaces)
-    for path in package_paths:
-        packages[path] = parse_package(os.path.join(basepath, path), warnings=warnings)
-    return packages
+    parser = _PackageParser(capture_warnings=isinstance(warnings, list))
+    results = multiprocessing.Pool(4).map(parser, package_paths)
```
dirk-thomas: Should we use something more "dynamic" like `multiprocessing.cpu_count()` as a default (and fallback to a hard coded number like `4`)?

mikepurvis: Looks like you can just leave off the number and it will fall back on cpu_count as the default. Will change (I found empirically that 4 worked well for me).

src/catkin_pkg/packages.py
mikepurvis: Yeah, probably— will change.

src/catkin_pkg/packages.py
mikepurvis: Yeah, can do. I find the tuple unpacking a bit more readable than explicit indexing, but it could go either way.

mikepurvis: Okay, I've dropped the comprehensions in favour of zipping; I think that's maybe a bit clearer.

dirk-thomas:
This looks good.

It might be a bit out of scope for the patch but it would be great if the API would allow the same improvement for this use (https://github.com/ros/rosdistro/blob/master/test/test_build_caches.py#L40-L41) which passes the xml rather than the paths.

dirk-thomas:
I tried to apply the same idea to the `parse_packages_by_strings` case. But it seems since in that case I already have all the xml in memory and only parse the xml in parallel it doesn't improve the performance (it actually is about 10% slower).

I tried calling the API `find_packages_allowing_duplicates` locally pointing to different folders. For smaller numbers (e.g. 100) the patched version was slower. For larger cases (e.g. 1000) it was certainly faster. Should be switch between the old and new behavior based on the number of packages or just optimize for the large scale? @ros-infrastructure/ros_team Any opinions?

mikepurvis:
There's certainly a cost to summoning the other processes. I just wasn't sure if the cost was enough to justify switching between implementations, but happy to go either way on it.

mikepurvis:
I don't really have the time/energy for it at the moment, but it would be interesting to study whether the time in the kind of situation is predominantly being spent in syscalls, XML parsing, or Python somewhere. If it's either of the first two, then perhaps there are cases where the GIL isn't being properly released? If that was the case, and it could be fixed, then you'd definitely expect a threaded implementation to be the fastest, particularly since the setup cost then is way smaller.

mikepurvis:
Maybe there's a cleaner implementation here where multiprocessing is only used to get them into memory in parallel, and then they're parsed serially in a `parse_package_strings` function. That neatly sidesteps the matter of handling `warnings` properly, since that portion of things goes back to being serial.

dirk-thomas:
I did some more benchmarking with a path containing ~1000 packages:

* `find_package_paths` takes about 1/3 of a second
* `get_package_xml` for all packages takes only 10-15ms (independent of using / not using  `multiprocessing`)
* `parse_package_string` takes 3/4s (without), 1/4s (with `multiprocessing`)

Therefore I have created an updated PR which uses `multiprocessing` only for the xml parsing part: see #171. 

With this implementation the performance with `multiprocessing` is only worse for < 10 packages. And the overhead is only around 10ms. Therefore I didn't implement any conditional logic to skip using `multiprocessing` in case of few packages.

mikepurvis:
Interesting! I admittedly didn't dig in, but it definitely looked to me like the loading rather than parsing was the more expensive piece. Thanks for verifying that not to be the case.

dirk-thomas:
Maybe it also depends on what kind of system your are using? Does the new PR show the same improvements for your use case as this one?

PR #171, opened 2017-05-16
use multiprocessing when parsing packages (take 2)

dirk-thomas:
Replaces #170.


dirk-thomas:
Fixed the patch to pass the tests...

dirk-thomas:
@mikepurvis Maybe you can try this patch and comment if it works and performs in your case as desired?

PR #182, opened 2017-06-27
Added dashes to catkin packages naming

yotabits:
[Rep 127](http://www.ros.org/reps/rep-0127.html#name) seems to allow dashes,
On my [last released package](https://github.com/ros/rosdistro/pull/15095), following @dirk-thomas guidance, the naming has been modified from python_watchdog to python-watchdog.

Now I'm getting plenty of warnings because of the name used when I run of build code.

@mikepurvis Why did you make a special case for catkin packages on https://github.com/ros-infrastructure/catkin_pkg/pull/162.

Maybe I'm missing something, please highlight me :)

### DIFF
--- a/test/test_package.py
+++ b/test/test_package.py
@@ -158,11 +158,6 @@ class PackageTest(unittest.TestCase):
         pack.validate(warnings=warnings)
         self.assertIn('naming conventions', warnings[0])
 
-        pack.name = 'bar-bza'
-        warnings = []
-        pack.validate(warnings=warnings)
-        self.assertIn('naming conventions', warnings[0])
-
         pack.name = 'BAR'
         warnings = []
         pack.validate(warnings=warnings)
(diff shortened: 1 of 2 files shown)


mikepurvis:
The special case was because they used to not be allowed at all, and I lobbied to carve out an exception for only the case I needed— which was non-catkin packages.

However, I'm totally fine to see it extended.

mikaelarguedas:
I don't have authority here but I agree that given the amendment made to REP127/REP140 in https://github.com/ros-infrastructure/rep/pull/126 the validation should be updated accordingly. This will allow to suppress the warnings currently showing up for all packages using the same name as the debian package to:
"support maintaining a consistent dependency name when transitioning back and forth between a system dependency and in-workspace package"

test/test_package.py
```diff
@@ -158,11 +158,6 @@ def test_validate_package(self):
         pack.validate(warnings=warnings)
         self.assertIn('naming conventions', warnings[0])
 
-        pack.name = 'bar-bza'
-        warnings = []
-        pack.validate(warnings=warnings)
-        self.assertIn('naming conventions', warnings[0])
-
```
dirk-thomas: Instead of removing this test can you please update it to ensure that it is a valid package name now.

dirk-thomas:
@yotabits Any news on this?

dirk-thomas:
Closing this due to no response by the author and in favor of #199.

PR #183, opened 2017-07-05
fix handling of hidden folders

ipa-mdl:
The old code modified the `dirnames` list while iterating over it, which results in skipping some entries and sporadically hidden folders get returned.

Alternative implementation:
```python
for dirname in dirnames[:]:
    if dirname.startswith('.'):
        dirnames.remove(dirname)
```

### DIFF
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -66,7 +66,8 @@ def find_package_paths(basepath, exclude_paths=None, exclude_subspaces=False):
             paths.append(os.path.relpath(dirpath, basepath))
             del dirnames[:]
             continue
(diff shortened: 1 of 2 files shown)


ipa-mdl:
ready for review :)

ipa-mdl:
The test might pass anyway if `test3' gets listed first.

dirk-thomas:
Thank you for fixing this common bug. It is always surprising to me that this code snippet existed for 5+ years without anybody noticing :wink:

ipa-mdl:
> It is always surprising to me that this code snippet existed for 5+ years without anybody noticing

Most users don't need to hide packages inside their git repositories ;)

PR #178, opened 2017-06-02
fix encode/decode problems during changelog generation

dirk-thomas:
Replaces and fixes #177.

PR #184, opened 2017-07-05
Open changelogs as utf-8

nuclearsandwich:
Observed when generating packages for ROS 2. Tested manually on python 2.7.12 and 3.5.2.

There currently aren't any tests that exercise the changelog file IO. Would such tests be worth adding?

### DIFF
--- a/src/catkin_pkg/changelog.py
+++ b/src/catkin_pkg/changelog.py
@@ -53,6 +53,8 @@ import os
 import pkg_resources
 import re
 
+from io import open
+
 try:
     _unicode = unicode
 except NameError:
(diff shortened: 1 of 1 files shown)


src/catkin_pkg/changelog.py
```diff
@@ -53,6 +53,8 @@
 import pkg_resources
 import re
 
+from io import open
```
nuclearsandwich: This import is required so that the encoding kwarg is accepted in both python3 and python2.

mikaelarguedas: oh interesting that didnt seem to be needed in #178 would read as binary (`rb`) work without this import ?

nuclearsandwich: Is there a reason to prefer decoding after the fact other than that's what's already done elsewhere? As far as I can tell we never _want_ to work with the unencoded file contents.

dirk-thomas:
Can you please describe a reproducible example (or even better add a test) which fails without and passes with this patch.

dirk-thomas:
@nuclearsandwich ping

nuclearsandwich:
When I went back to isolate and reproduce this issue I wasn't able to do so. Unless/until it pops up again this is not needed.

PR #190, opened 2017-08-08
chroot environment does not suport multithreading

k-okada:
on some environment, python multithreading raise error with OSError38
(https://stackoverflow.com/questions/6033599/oserror-38-errno-38-with-multiprocessing)
In my case, I did chroot my environment and had this issue,

I'm not sure what is the best way to avoid this problem, here is a workaround by calling Lock() function and raise error when it is not available.

```
ERROR: Rosdep experienced an error: [Errno 38] Function not implemented
Please go to the rosdep page [1] and file a bug report with the stack trace belo
w.
[1] : http://www.ros.org/wiki/rosdep

rosdep version: 0.11.8

  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 137, in rosdep_m
ain
    exit_code = _rosdep_main(args)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 383, in _rosdep_
main
    return _package_args_handler(command, parser, options, args)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 458, in _package
_args_handler
    pkgs = find_catkin_packages_in(path, options.verbose)
  File "/usr/lib/python2.7/dist-packages/rosdep2/catkin_packages.py", line 33, i
n find_catkin_packages_in
    packages = find_packages(path)
  File "/usr/lib/python2.7/dist-packages/catkin_pkg/packages.py", line 87, in find_packages
    packages = find_packages_allowing_duplicates(basepath, exclude_paths=exclude_paths, exclude_subspaces=exclude_subspaces, warnings=warnings)
  File "/usr/lib/python2.7/dist-packages/catkin_pkg/packages.py", line 142, in find_packages_allowing_duplicates
    pool = multiprocessing.Pool()
  File "/usr/lib/python2.7/multiprocessing/__init__.py", line 232, in Pool
    return Pool(processes, initializer, initargs, maxtasksperchild)
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 138, in __init__
    self._setup_queues()
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 234, in _setup_queues
    self._inqueue = SimpleQueue()
  File "/usr/lib/python2.7/multiprocessing/queues.py", line 352, in __init__
    self._rlock = Lock()
  File "/usr/lib/python2.7/multiprocessing/synchronize.py", line 147, in __init__
    SemLock.__init__(self, SEMAPHORE, 1, 1)
  File "/usr/lib/python2.7/multiprocessing/synchronize.py", line 75, in __init__
    sl = self._semlock = _multiprocessing.SemLock(kind, value, maxvalue)
OSError: [Errno 38] Function not implemented
```

### DIFF
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -138,18 +138,7 @@ def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subs
             # https://stackoverflow.com/questions/6033599/oserror-38-errno-38-with-multiprocessing
             parallel = False
 
-    if parallel:
-        # use multiprocessing pool
-        parser = _PackageParser(warnings is not None)
-        try:
-            path_parsed_packages, warnings_lists = zip(*pool.map(parser, data))
-        finally:
-            pool.close()
-            pool.join()
-        if parser.capture_warnings:
-            map(warnings.extend, warnings_lists)
-        return dict(path_parsed_packages)
-    else:
+    if not parallel:
         #  use sequential loop
         parsed_packages = {}
         for xml, path, filename in data:
@@ -157,6 +146,17 @@ def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subs
                 xml, filename=filename, warnings=warnings)
             parsed_packages[path] = parsed_package
         return parsed_packages
+
+    # use multiprocessing pool
+    parser = _PackageParser(warnings is not None)
+    try:
+        path_parsed_packages, warnings_lists = zip(*pool.map(parser, data))
+    finally:
+        pool.close()
+        pool.join()
+    if parser.capture_warnings:
+        map(warnings.extend, warnings_lists)
+    return dict(path_parsed_packages)
 
 
 def verify_equal_package_versions(packages):
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -132,7 +132,6 @@ def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subs
     parallel = len(data) > 100
     if parallel:
         try:
-            parser = _PackageParser(warnings is not None)
             pool = multiprocessing.Pool()
         except OSError:
             # On chroot environment, multiprocessing is not available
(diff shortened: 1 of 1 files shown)


src/catkin_pkg/packages.py
```diff
@@ -129,25 +129,27 @@ def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subs
...
+        pool = multiprocessing.Pool()
+        try:
+            path_parsed_packages, warnings_lists = zip(*pool.map(parser, data))
+        finally:
+            pool.close()
+            pool.join()
+        if parser.capture_warnings:
+            map(warnings.extend, warnings_lists)
+        return dict(path_parsed_packages)
+    except:
```
dirk-thomas: Try except the whole code block seems to be excessive as well as excepting on any kind of exception. Please identify a specific function which can be tested for to choose between the two approaches.

k-okada: Thanks for comment, so you mean something like
```
def find_packages_allowing_duplicates():
   ...

 if len(data) < 100:
   return find_packages_allowing_duplicates_no_theads()

parser = _PackageParser(warnings is not None)
try:
    pool = multiprocessing.Pool()
except (OSError):
   return find_packages_allowing_duplicates_no_theads()
try:
 path_parsed_packages, warnings_lists = zip(*pool.map(parser, data))
...
```
??

dirk-thomas: Roughly yes. But you don't need to move part of the logic into a function in order to do that. I would try to do a conditional logic like this:

```
parallel = len(data) > 100
if parallel:
    try:
        pool = multiprocessing.Pool()
    except OSError:  # is this the only possible exception?
        parallel = False

if parallel:
    # use multiprocessing pool
else:
    # use sequential loop
```

k-okada:
thank you for comment, I have updated the branch.

--
◉ Kei Okada


2017-08-10 1:27 GMT+09:00 Dirk Thomas <notifications@github.com>:

src/catkin_pkg/packages.py
```diff
@@ -129,25 +129,35 @@ def find_packages_allowing_duplicates(basepath, exclude_paths=None, exclude_subs
     if not data:
         return {}
 
-    if len(data) < 100:
+    parallel = len(data) > 100
+    if parallel:
+        try:
+            parser = _PackageParser(warnings is not None)
```
dirk-thomas: This line shouldn't raise and therefore shouldn't be in the try/except block. It should be moved in to `parallel` block below.

k-okada: ok, updated

dirk-thomas:
Thank you for the improvement.

PR #193, opened 2017-10-11
added XML namespace support

ipa-mdl:
This patch relaxes the checking of unknown attributes in `package.xml` to support [XML namespaces](https://www.w3.org/TR/xml-names/).
`xml.dom.minidom` already checks the validity of the prefixes (see test).

Use cases:
* provide schema in `xsi:schemaLocation` (not all parsers support [`xml-model`](https://www.w3.org/TR/xml-model/#dt-xml-model))
* XMI data in `<export>`
* tag meta-data annotations, e.g. for generators or cross-linking

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -419,6 +419,15 @@ def parse_package(path, warnings=None):
         raise
 
 
+def _check_known_attributes(node, known):
+    if node.hasAttributes():
+        attrs = map(str, node.attributes.keys())
+        # colon is the namespace separator in attributes, xmlns can be added to any tag
+        unknown_attrs = [attr for attr in attrs if not (attr in known or attr == 'xmlns' or ':' in attr)]
+        if unknown_attrs:
+            return ['The "%s" tag must not have the following attributes: %s' % (node.tagName, ', '.join(unknown_attrs))]
+    return []
+
 def parse_package_string(data, filename=None, warnings=None):
     """
     Parse package.xml string contents.
@@ -539,9 +548,7 @@ def parse_package_string(data, filename=None, warnings=None):
         pkg.exports = exports
 
     # verify that no unsupported tags and attributes are present
-    unknown_root_attributes = [attr for attr in root.attributes.keys() if str(attr) != 'format']
-    if unknown_root_attributes:
-        errors.append('The "package" tag must not have the following attributes: %s' % ', '.join(unknown_root_attributes))
+    errors += _check_known_attributes(root, ['format'])
     depend_attributes = ['version_lt', 'version_lte', 'version_eq', 'version_gte', 'version_gt']
     known = {
         'name': [],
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Does this require and update to REP [127](http://www.ros.org/reps/rep-0127.html) and [141](http://www.ros.org/reps/rep-0141.html)?

Do manifest files using these attributes validate using the current schemas (format [1](http://download.ros.org/schema/package_format1.xsd) and [2](http://download.ros.org/schema/package_format2.xsd))?

ipa-mdl:
> Does this require and update to REP 127 and 141?

Not necessarily, namespaces are more or less an orthogonal XML feature.
There is a section on [XML restrictions](http://www.ros.org/reps/rep-0127.html#xml-syntax-restrictions), but this does not conflict with the usage of namespaces.
This could be stated explicitly, but I don't think that this is needed.

> Do manifest files using these attributes validate using the current schemas (format 1 and 2)?

Yes, I have tested it with a patched version of the [example in the wiki](http://wiki.ros.org/catkin/package.xml#Required_Tags):
```xml
<package format="2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <name>foo_core</name>
  <version xsi:type="VersionType">1.2.4</version>
  <description>
  This package provides foo capability.
  </description>
  <maintainer email="ivana@osrf.org">Ivana Bildbotz</maintainer>
  <license>BSD</license>
</package>
```
[example.xml.txt](https://github.com/ros-infrastructure/catkin_pkg/files/1377256/example.xml.txt)

```
$ xmllint --noout --schema  http://download.ros.org/schema/package_format2.xsd example.xml.txt 
example.xml.txt validates
```

ipa-mdl:
It might make sense to add this example as an addtional unit test.

dirk-thomas:
A parser which can read manifests currently specified by the REP might choke on `xsi:type="VersionType"` since it doesn't expect it to be allowed. I am always hesitant to change a spec like this after the fact. There *might* be code out there which break - there is just no way of knowing. We could either amend the REPs 127 and 140 to explicitly mention XML namespaces or only allow them in an upcoming format 3.

@ros-infrastructure/ros_team What do others think about this?

ipa-mdl:
The example was just given for the validation, `xsi:type="VersionType"` does not make much sense.
I'd like to use these things in the `<export>` section, but the namespace might get defined in the root element in order to be references later.
It should not break any parser.

dirk-thomas:
> It should not break any parser.

"Should". Just to make an example: someone might have implemented a custom parser to fit on a micro controller and it will simply fail if the root element has an attribute not specified in the REP. I just want to point out that we are aware of the possibility and a change in the REP would have that risk of breakage.

wjwwood:
tl;dr +0, I don't see why we need it exactly, but the potential harm I think is unlikely to occur.

I don't see the impetus for allowing the XML namespaces (I'm sure there are use cases, but I haven't seen a lot of people asking for it). But on the other hand I think it's unlikely very many people are using a non-compliant XML parser to interpret the manifest files, and if they are, we did declare to be using XML, so it would be a bad assumption that we don't use a feature of the specification. And in that case I think it's fine for custom parsers to be broken and need to switch to a compliant parser or update their "custom" parser to allow or ignore these attributes.

dirk-thomas:
Sounds good to me.

Thank you for the patch and thanks for the discussion / feedback.

PR #195, opened 2017-12-06
[WIP] Support for parsing package format 3

nuclearsandwich:
🚧 WIP. Do not merge. 🚧 

This PR adds the minimum changes necessary to run bloom, and soon the ROS 2 buildfarm supporting the package format 3 from in https://github.com/ros-infrastructure/rep/pull/138

At the moment the goal of this branch is to successfully parse format 3 packages without errors and is not intended to intended to bring complete support for format 3.

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -454,7 +454,7 @@ def parse_package_string(data, filename=None, warnings=None):
     # format attribute
     value = _get_node_attr(root, 'format', default=1)
     pkg.package_format = int(value)
-    assert pkg.package_format in [1, 2], "Unable to handle package.xml format version '%d', please update catkin_pkg (e.g. on Ubuntu/Debian use: sudo apt-get update && sudo apt-get install --only-upgrade python-catkin-pkg)" % pkg.package_format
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Now that the release crunch is over we can examine this in more detail.

Should catkin_pkg implement support for the draft format 3? If so, I can start updating this PR with changes backported from ament_package.

dirk-thomas:
> Should catkin_pkg implement support for the draft format 3?

The goal of REP 149 is to allow a ROS package to be build with ROS 1 as well as ROS 2 from a single source. So yes, `catkin_pkg` as well as other tools in the ROS 1 toolchain should support format 3.

tspicer01:
Do we have an ETA on when we may have this support ready?
We've just moved our ROS2 builds from beta3 to GA and some of our legacy tooling that we use for both ROS and ROS2 fail on the packages that have been moved to format3.

dirk-thomas:
> Do we have an ETA on when we may have this support ready?

I have created #197 to implement full REP 149 support. In the current state it does successfully parse all manifest files of ROS 2. It doesn't utilize the new information yet though, e.g. group dependencies or dependency conditions.

---

I think we should keep this around since it was used to create the releases into ardent. It will be convenient for future patch releases to not require changing the process.

dirk-thomas:
@nuclearsandwich Can you please try if #197 is sufficient for releasing ROS 2 packages. If yes, this can be closed.

nuclearsandwich:
This PR is superceded by #197

PR #197, opened 2018-01-03
support package format 3

dirk-thomas:
As defined in [REP 149](http://www.ros.org/reps/rep-0149.html).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ python:
   - "3.3"
 # command to install dependencies
 install:
-  - pip install nose coverage argparse python-dateutil docutils
+  - pip install nose coverage argparse python-dateutil docutils pyparsing
 # command to run tests
 script:
   - nosetests -s --tests test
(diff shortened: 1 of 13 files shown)


dirk-thomas:
In the current state it does successfully parse all manifest files of ROS 2. It doesn't utilize the new information yet though, e.g. group dependencies or dependency conditions.

dirk-thomas:
The patch now does utilize the new information in the topological order as well as uses the context to evaluate dependency conditions.

Ready for review.

dirk-thomas:
> I believe the pyparsing dependency needs to be added to the `stdeb.cfg` file too.

Thank you - good catch. I added it [here](https://github.com/ros-infrastructure/catkin_pkg/pull/197/commits/92ee7f2a36eba59f5b066865e99a3ca79dd7af8c#diff-419aa3e2c65a318d41b25af356c30f36).

dirk-thomas:
Waiting for a [confirmation](https://github.com/ros-infrastructure/catkin_pkg/pull/195#issuecomment-355404258) from @nuclearsandwich before merging...

PR #199, opened 2018-01-17
allow dash in catkin package names

dirk-thomas:
Replaces #182.

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -222,16 +222,10 @@ class Package(object):
         if not re.match('^[a-zA-Z0-9][a-zA-Z0-9_-]*$', self.name):
             errors.append('Package name "%s" does not follow naming conventions' % self.name)
         else:
-            if self.get_build_type() == 'catkin':
-                if not re.match('^[a-z][a-z0-9_]*$', self.name):
-                    new_warnings.append(
-                        'Catkin package name "%s" does not follow the naming conventions. It should start with '
-                        'a lower case letter and only contain lower case letters, digits, and underscores.' % self.name)
-            else:
-                if not re.match('^[a-z][a-z0-9_-]*$', self.name):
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
@dirk-thomas wrote (in #182):
> I just hope we will not run into the problem of duplicate package names afterwards or uncertainty about how the CMake variables are being exported (since that was a main reason why this rule enforced uniqueness and consistency).

can you say anything about the exported CMake variables? Would the dashes be retained there?

dirk-thomas:
> can you say anything about the exported CMake variables? Would the dashes be retained there?

CMake variables can't contain dashes.

gavanderhoorn:
> CMake variables can't contain dashes.

exactly, so will the variables for those pkgs have underscores instead then? Is that the potential confusion you are referring to?

dirk-thomas:
> so will the variables for those pkgs have underscores instead then? Is that the potential confusion you are referring to?

Yes, I would assume using underscores would be a reasonable mapping.

A package name with dashes is more common for e.g. Python packages though. I hope this won't be used for CMake packages often.

dirk-thomas:
@ros-infrastructure/ros_team Please review.

wjwwood:
> CMake variables can't contain dashes.

Actually that's not true (/shudder) 🤢:

```
william@johnnycake /tmp
% cat test.cmake
set(a-dash "Hello World!")
message(STATUS ${a-dash})

william@johnnycake /tmp
% cmake -P test.cmake
-- Hello World!
```

dirk-thomas:
> > CMake variables can't contain dashes.

> Actually that's not true (/shudder) :nauseated_face::

Interesting, that is the first time I have seen that but the [docs](https://cmake.org/cmake/help/v3.0/manual/cmake-language.7.html#variables) "agree" with it.

PR #202, opened 2018-02-13
Support file attribution of license tag

otamachan:
Implement https://github.com/ros-infrastructure/rep/pull/157 .

Just for reference because the proposal is now under review.

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -103,6 +103,7 @@ class Package(object):
                         slot.append(deepcopy(d))
             del kwargs['run_depends']
         self.filename = filename
+        self.licenses = [l if isinstance(l, License) else License(l) for l in self.licenses]
         # verify that no unknown keywords are passed
         unknown = set(kwargs.keys()).difference(self.__slots__)
         if unknown:
@@ -370,6 +371,14 @@ class Export(object):
         return txt
 
 
+# Subclassing ``str`` to keep backward compatibility.
+class License(str):
+    def __new__(cls, value, file_=None):
+        obj = str.__new__(cls, str(value))
+        obj.file = file_
+        return obj
+
+
 class Person(object):
     __slots__ = ['name', 'email']
 
@@ -553,7 +562,10 @@ def parse_package_string(data, filename=None, warnings=None):
     # at least one license
     licenses = _get_nodes(root, 'license')
     for node in licenses:
-        pkg.licenses.append(_get_node_value(node))
+        pkg.licenses.append(License(
+            _get_node_value(node),
+            _get_node_attr(node, 'file', default=None)
+        ))
 
     errors = []
     # dependencies and relationships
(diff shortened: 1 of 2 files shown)


src/catkin_pkg/package.py
```diff
@@ -370,6 +371,14 @@ def __str__(self):
         return txt
 
 
+# Subclassing ``str`` to keep backward compatibility.
+class License(str):
+    def __new__(cls, value, path=''):
```
dirk-thomas: The default value should be `None` in order to be distinguishable from an existing attribute with an empty string value.

src/catkin_pkg/package.py
```diff
@@ -370,6 +371,14 @@ def __str__(self):
...
+        obj = str.__new__(cls, str(value))
+        obj.path = path
```
dirk-thomas: I would suggest to use `file` rather then `path` to follow the naming in XML.

dirk-thomas:
Please also rebase your branch so that the PR becomes mergable.

src/catkin_pkg/package.py
otamachan: I will change the default value is None.

src/catkin_pkg/package.py
otamachan: ``file`` is a builtin function. But for this case, this will not cause a problem. I will use ``file`` instead of ``path`

dirk-thomas: I think the class member should `file` (so `obj.file = ...`). But since `file` is a builtin in Python the argument of the function should be `file_` (so `file_=None`).

otamachan: OK. Fixed.

src/catkin_pkg/package_templates.py
```diff
@@ -363,7 +365,10 @@ def create_package_xml(package_template, rosdistro, meta=False):
 
     licenses = []
     for plicense in package_template.licenses:
-        licenses.append('  <license>%s</license>\n' % plicense)
+        if package_template.package_format is not None and int(package_template.package_format) > 2:
+            licenses.append('  <license file="%s">%s</license>\n' % (plicense.file, plicense))
```
dirk-thomas: Does this result in generating an attribute with `file="None"`? If yes, this should skip the xml attribute in the case the value is None instead.

otamachan: Yes. So I will skip the generation of the attribution if the value is None.

But I think it's better to have this attribution for many licenses, I would like to generate an empty attribution by default.

So is it ok to change
from
```
        licenses = list(licenses or ["TODO"])
```
to
```
        licenses = list(licenses or [License("TODO", "")])
```
?
This might affect the ``bloom`` releasing(https://github.com/ros-infrastructure/bloom/pull/470/files#diff-9a801db47dfcf7ec0e97334dcb47b796R425)

dirk-thomas: I see you point that not having the attribute will likely result in many people not using it. The opposite downside is that an empty attribute isn't valid (according to the REP as well as the `bloom` logic you referenced). Therefore I don't think that is an option either.

otamachan: I will drop the commit(https://github.com/ros-infrastructure/catkin_pkg/pull/202/commits/104e6ceef352719e82285b4774a8b810347373bd)

dirk-thomas:
Thank you for working on this.

PR #203, opened 2018-02-14
Respect Git .mailmap file for generated changelog

KindDragon:


KindDragon:
https://git-scm.com/docs/git-log
```
%an: author name
%aN: author name (respecting .mailmap, see git-shortlog[1] or git-blame[1])
```

dirk-thomas:
Thank you for the patch.

PR #191, opened 2017-09-01
catkin_create_pkg use package.xml format 2

scpeters:
This is a work-in-progress to resolve #96. I tried updating it to match the new tags, but I haven't tested it yet.

PR #205, opened 2018-02-22
Replace run_depend

csukuangfj:
Replace run_depend with exec_depend described in http://docs.ros.org/jade/api/catkin/html/howto/format2/migrating_from_format_1.html


dirk-thomas:
Follow up of #191.

Thank you for the patch.

PR #209, opened 2018-03-22
InvalidPackage with package name

Nickolaim:
`parse_package_string()` raises `InvalidPackage` exception with the package name in the message.

This work is needed for [https://github.com/ros-infrastructure/rosdep/issues/598](https://github.com/ros-infrastructure/rosdep/issues/598)

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -677,7 +677,7 @@ def _get_nodes(parent, tagname):
 def _get_node(parent, tagname, filename):
     nodes = _get_nodes(parent, tagname)
     if len(nodes) != 1:
-        raise InvalidPackage('The manifest must contain exactly one "%s" tags' % tagname, filename)
+        raise InvalidPackage('The manifest must contain exactly one "%s" tag' % tagname, filename)
     return nodes[0]
 
 
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -470,11 +470,7 @@ def parse_package(path, warnings=None):
     :raises: :exc:`IOError`
     """
     xml, filename = _get_package_xml(path)
-    try:
-        return parse_package_string(xml, filename, warnings=warnings)
-    except InvalidPackage as e:
-        e.args = ['Invalid package manifest "{0}": {1}'.format(filename, e)]
-        raise
+    return parse_package_string(xml, filename, warnings=warnings)
 
 
 def _check_known_attributes(node, known):
@@ -486,6 +482,7 @@ def _check_known_attributes(node, known):
             return ['The "%s" tag must not have the following attributes: %s' % (node.tagName, ', '.join(unknown_attrs))]
     return []
 
+
 def parse_package_string(data, filename=None, warnings=None):
     """
     Parse package.xml string contents.
@@ -496,6 +493,14 @@ def parse_package_string(data, filename=None, warnings=None):
     :returns: return parsed :class:`Package`
     :raises: :exc:`InvalidPackage`
     """
+    try:
+        return _parse_package_string(data, filename=filename, warnings=warnings)
+    except InvalidPackage as e:
+        e.args = ('Invalid package manifest "{0}": {1}'.format(filename, e),)
+        raise
+
+
+def _parse_package_string(data, filename=None, warnings=None):
     try:
         root = dom.parseString(data)
     except Exception as ex:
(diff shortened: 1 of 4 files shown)


dirk-thomas:
Instead of embedding the filename into every error message (the current patch only does it in a few places) wouldn't it make more sense to store the filename as well as the list of errors in the `InvalidPackage` instance as members? Then the string representation can combine them in a standard way.

Nickolaim:
Yep, it makes sense, will do.  Please note that file name is not always available at the moment when the exception is raised, so the code for populating file name in some exception handlers will remain.
________________________________

src/catkin_pkg/package.py
```diff
@@ -421,7 +414,14 @@ def parse_package_for_distutils(path=None):
...
 
 class InvalidPackage(Exception):
-    pass
+    def __init__(self, msg, package_path=None):
+        self.msg = msg
+        self.package_path = package_path
+        Exception.__init__(self, self.msg)
+
+    def __str__(self):
+        result = '' if not self.package_path else 'Error(s) in package \'%s\':\n' % self.package_path
```
dirk-thomas: Please avoid escaping in the string. To using single quotes within the string use double quotes for the string itself:

```
"Error(s) in package '%s':\n"
```

src/catkin_pkg/package.py
```diff
@@ -503,22 +503,23 @@ def parse_package_string(data, filename=None, warnings=None):
     try:
         return _parse_package_string(data, filename=filename, warnings=warnings)
     except InvalidPackage as e:
-        e.args = ('Invalid package manifest "{0}": {1}'.format(filename, e),)
+        if not e.package_path:
+            e.package_path = filename
```
dirk-thomas: In which case would the raise exception not have a package path set?

test/test_package.py
```diff
@@ -261,6 +261,18 @@ def test_validate_package(self):
...
+    def test_invalid_package_exception(self):
+        try:
+            raise InvalidPackage('foo')
+        except InvalidPackage as e:
+            self.assertEqual('foo', str(e))
+            self.assertEqual(None, e.package_path)
+        try:
+            raise InvalidPackage('foo', package_path='\\bar')
+        except InvalidPackage as e:
+            self.assertEqual('Error(s) in package \'\\bar\':\nfoo', str(e))
```
dirk-thomas: Same here regarding escaping the quotes.

test/test_package.py
```diff
@@ -261,6 +261,18 @@ def test_validate_package(self):
...
             dep_type.remove(depend)
 
+    def test_invalid_package_exception(self):
+        try:
+            raise InvalidPackage('foo')
+        except InvalidPackage as e:
+            self.assertEqual('foo', str(e))
+            self.assertEqual(None, e.package_path)
+        try:
+            raise InvalidPackage('foo', package_path='\\bar')
```
dirk-thomas: Rather then using an escaped backslash I would suggest to using simpler like `/some/path` since it doesn't even have to exist.

src/catkin_pkg/package.py
Nickolaim: Fixed.

test/test_package.py
Nickolaim: Fixed.

test/test_package.py
Nickolaim: Fixed.

src/catkin_pkg/package.py
Nickolaim: [_get_node() and _get_optional_node()](https://github.com/Nickolaim/catkin_pkg/blob/f63344620230ce72d45d84ade2cbf09a299cf886/src/catkin_pkg/package.py#L686-L697) that e.g. called from [L#534](https://github.com/Nickolaim/catkin_pkg/blob/f63344620230ce72d45d84ade2cbf09a299cf886/src/catkin_pkg/package.py#L534) and [#L616](https://github.com/Nickolaim/catkin_pkg/blob/f63344620230ce72d45d84ade2cbf09a299cf886/src/catkin_pkg/package.py#L616)

test/test_package.py
Nickolaim: BTW maybe there is a way to have a flake8 check for cases like this, so human spend time on more important stuff.

dirk-thomas: In ROS 2 we have these kind of checks in place. Not for long existing ROS 1 packages though at the moment...

src/catkin_pkg/package.py
dirk-thomas: Can you update the patch to catch, update and raise the exception around those two calls. Then there is no reason for the handling here and `_parse_package_string` could be eliminated all together and the implementation can move back into `parse_package_string` directly.

Nickolaim: Done.

test/test_package.py
Nickolaim: I have PR #211  that at some level addresses this issue.  Happy to bring other checks from ament_flake8 later.  Don't want to do it all in one change - fixing a lot of flake8 issues is hard to review.

dirk-thomas:
Thank you for working on this.

PR #215, opened 2018-05-21
target newer Python versions

dirk-thomas:
Connect to ros/rosdistro#17903.


dirk-thomas:
While not actively testing for it we keep Python 3.2 in the Debian package until precise is being dropped.

PR #217, opened 2018-05-30
flake8-docstrings

Nickolaim:
Bring checks for flake8-docstrings; fix the code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ python:
 install:
   - pip install nose coverage argparse python-dateutil docutils pyparsing
   - pip install flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated
+  - pip install flake8-docstrings
 # command to run tests
 script:
   - nosetests -s --tests test
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,7 @@ python:
 # command to install dependencies
 install:
   - pip install nose coverage argparse python-dateutil docutils pyparsing
-  - pip install flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated
-  - pip install flake8-docstrings
+  - pip install flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated flake8-docstrings
 # command to run tests
 script:
   - nosetests -s --tests test
--- a/src/catkin_pkg/cmake.py
+++ b/src/catkin_pkg/cmake.py
@@ -47,7 +47,7 @@ def get_metapackage_cmake_template_path():
     return os.path.join(os.path.dirname(__file__), rel_path)
 
 
-def configure_file(template_file, environment):  # noqa: ignore=D402
+def configure_file(template_file, environment):  # noqa: D402
     """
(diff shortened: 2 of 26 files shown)


src/catkin_pkg/cmake.py
```diff
@@ -38,33 +38,33 @@
...
+    """
     rel_path = os.path.join('templates', 'metapackage.cmake.in')
     return os.path.join(os.path.dirname(__file__), rel_path)
 
 
 def configure_file(template_file, environment):
-    '''
-    Evaluate a .in template file used in CMake with configure_file().
+    """
+    Evaluate a .in template file used in CMake with configure_string().
```
dirk-thomas: Can you please explain why you change this?

.travis.yml
```diff
@@ -10,6 +10,7 @@ python:
 install:
   - pip install nose coverage argparse python-dateutil docutils pyparsing
   - pip install flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated
+  - pip install flake8-docstrings
```
dirk-thomas: Please insert this package into the previous line which already installs all flake8 plugins.

Nickolaim: Done.  Please note that I plan to add 2 more packages, and the line will be even longer.

src/catkin_pkg/cmake.py
Nickolaim: The original comment triggered `D402 First line should not be the function's "signature"` issue.

The description  does not make sense to me - _`configure_file()` is doing `Evaluate .. with configure_file()`_.

Here is the function body:
```
    with open(template_file, 'r') as f:
        template = f.read()
        return configure_string(template, environment)
```

I assumed that there is a typo in the original comment, and replaced `configure_file()` with `configure_string()`.

The alternative is to revert the change and to suppress D402.

dirk-thomas: This Python function has the same functionality as CMakes `configure_file` function. So the comment should stay as is and the warning can be suppressed.

Nickolaim: Fixed. 
Note - the suppression says 'Ignore only D402'; in reality, all issues in the function definition and the comment are suppressed. I was not able to find a way to suppress the issue for the comment only.

Nickolaim:
Anything else that needs to be addressed in this PR?

src/catkin_pkg/cmake.py
```diff
@@ -38,17 +38,17 @@
...
     :returns: ``str`` location of the metapackage CMakeLists.txt CMake template
-    '''
+    """
     rel_path = os.path.join('templates', 'metapackage.cmake.in')
     return os.path.join(os.path.dirname(__file__), rel_path)
 
 
-def configure_file(template_file, environment):
-    '''
+def configure_file(template_file, environment):  # noqa: ignore=D402
```
dirk-thomas: The reason why this comment suppresses all warnings is the way it is written. Please use `# noqa: D402` to only suppress that specific warning. That style ensures that other warnings are not suppressed.

Nickolaim: Fixed.  

There are some other incorrectly suppressed warnings in this project, I will address them in a separate PR.

dirk-thomas:
Thank you for your effort on improving the style!

PR #218, opened 2018-06-09
Fix minor code issues

Nickolaim:
Correctly suppress flake8 warnings
Update documentation, so data type is placed correctly


dirk-thomas:
Thanks again!

PR #219, opened 2018-06-10
Fix flake8-import-order issues

Nickolaim:
I100 Import statements are in the wrong order. 'import os' should be before 'import docutils.core' and in a different group.
I101 Imported names are in the wrong order. Should be PackageTemplate, create_package_files
I201 Missing newline between import groups. 'import argparse' is identified as Stdlib and 'from __future__ import print_function' is identified as Future.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -4,3 +4,4 @@ max-line-length = 200
 max-complexity = 54
 show-source = true
 statistics = true
+import-order-style = google
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,7 @@
 [flake8]
 ignore = C403,C404,C405,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202
-max-line-length = 200 
+import-order-style = google
 max-complexity = 54
+max-line-length = 200
 show-source = true
 statistics = true
-import-order-style = google
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
 # command to install dependencies
 install:
   - pip install nose coverage argparse python-dateutil docutils pyparsing
-  - pip install flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated flake8-docstrings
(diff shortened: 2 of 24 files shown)


mikaelarguedas:
change looks good to me. I don't know what strategy maintainers of this package prefer to adopt.
In ROS 2.0 we follow a case insensitive import order, strategy also named `googlestyle` in [flake8](https://github.com/ros-infrastructure/catkin_pkg/pull/219/files#diff-3aef618374a1b6377f71e2ee05fcf013). Using that style here would make the change of this PR minimal for files like [this one](https://github.com/ros-infrastructure/catkin_pkg/pull/219/files#diff-3aef618374a1b6377f71e2ee05fcf013L7) where the import order is already complying with the googlestyle import order.

Nickolaim:
Apparently changing default `import-order-style`is problematic.  Setting `import_order_style='google'` while calling `get_style_guide()` has no effect - when plugin is initialized, it uses only settings from command line and configuration file - details are in [flake8/legacy.py](https://gitlab.com/pycqa/flake8/blob/master/src/flake8/api/legacy.py#L36).

ROS2 has a solution for this problem - it passes options to `parse_configuration_and_cli()`, see [ament_flake8/main.py](https://github.com/ament/ament_lint/blob/master/ament_flake8/ament_flake8/main.py#L137).

In general, it does not mean that this is not possible to solve - moving configuration to a file is one way.  But this definitely out of the scope of this change.

mikaelarguedas:
Sounds good to me, thanks for giving it a shot :+1:

Nickolaim:
Any suggestions for this PR?

dirk-thomas:
When enabling `flake8-import-style` the same style should be preserved which the code currently already adheres to (at least mostly if not everywhere) which is the `google` style. This can e.g. be configured with `.flake8` file in the root of the package (e.g. https://github.com/ament/ament_package/blob/a6c73aeff77c188597211fee9f9e34508487b531/.flake8).

Nickolaim:
PR #222 moves flake8 configuration from the source code into .flake8 configuration file.  When it is merged, I update the settings to follow google import style.

Nickolaim:
Google import order style has been applied.

Nickolaim:
Both comments from the review are resolved.

dirk-thomas:
Thanks!

PR #222, opened 2018-06-17
Move flake8 configuration from the unit test into configuration file

Nickolaim:
Required for #219.

`max-complexity` is semi-random - it is the complexity of the `_main()` method + 1 for preventing more complex functions to be created.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,6 @@
+[flake8]
+ignore = C403,C404,C405,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202
+max-line-length = 200 
+max-complexity = 54
+show-source = true
+statistics = true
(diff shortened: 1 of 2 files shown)


.flake8
```diff
@@ -0,0 +1,6 @@
+[flake8]
+ignore = C403,C404,C405,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202
+max-line-length = 200 
+max-complexity = 54
```
dirk-thomas: Before `max_complexity=10` was passed. Why use a different value in the config file?

Nickolaim: 10 was the default value.  It did not work for the same reason setting import style to `google` does not work without this change - settings for plugins are NOT passed if they are set in API.  There are 2 ways to fix is - either move the settings to the config file OR update the code so it passes the dictionary to settings, as it is done in ros2 flake8.  References to the code are in [https://github.com/ros-infrastructure/catkin_pkg/pull/219#issuecomment-396092843](https://github.com/ros-infrastructure/catkin_pkg/pull/219#issuecomment-396092843).

PR #225, opened 2018-06-26
fix Python 3 compatibility in prepare_release module

### LINKED ISSUE #224: catkin_prepare_release does not work in Python3
It just says that there are uncommitted changes (even if there are not) and exits. I believe it has to do with subprocess returning bytes instead of a string in Python3, but I didn't have time to investigate further. I'm using the Python2 version for now. `catkin_generate_changelogs` may also have issues, but I was always using it from python2 because of when I update `catkin_pkg`.

dirk-thomas:
Fixes #224.


dirk-thomas:
I don't think this should / need to be a function. Also the check for the Python version isn't necessary. I would rather revert the last commit and apply the fix to the other locations.

PR #226, opened 2018-06-26
check for AMENT_IGNORE and COLCON_IGNORE files

dirk-thomas:
The current use case for this is `rosdep` which uses the `catkin_pkg` API to find packages. With the change any package ignored by either of the marker files will be skipped - no matter if it is `catkin`, `ament` or `colcon` specific.

### DIFF
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -44,7 +44,10 @@ def find_package_paths(basepath, exclude_paths=None, exclude_subspaces=False):
     """
     Crawls the filesystem to find package manifest files.
 
-    When a subfolder contains a file ``CATKIN_IGNORE`` it is ignored.
+    When a subfolder contains either of the following files it is ignored:
+    - ``AMENT_IGNORE``
+    - ``CATKIN_IGNORE``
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Lgtm as this fixes the current issue we have on development jobs.

It does seem a bit backward to have `catkin_pkg` have knowledge of all the build tools marker files hard-coded in its' code.
Was it considered to allow catkin_pkg consumers to specify marker file names? This would allow for more reuse of catkin_pkg and avoid hard coding external information in catkin_pkg itself.

dirk-thomas:
> Was it considered to allow catkin_pkg consumers to specify marker file names?

The same problem will apply to the caller, e.g. `rosdep`. So I don't think that would be an improvement.

mikaelarguedas:
Yeah I didn't mean it would be a silver bullet. Just that it seems that we're going in a direction of adding external knowledge to low level tools code that would ideally be caller-agnostic and not need to have any hard-coded knowledge of the arbitrary marker file the higher level tools define and use.

dirk-thomas:
:ship: in 0.4.5

PR #231, opened 2018-07-31
Fix permission denied issue on Windows

seanyen-msft:
The file created (and opened) by NamedTemporaryFile will be used in update_changelog_sections later and the second-time open() on the temp file will result in a permission denied on Windows because of Python implementation. The issue has been discussed here: https://bugs.python.org/issue14243

And one workaround talked about in the discussion is to pass "delete=False".

Here is the error messages for the test failure:

======================================================================
ERROR: test_update_changelog_unicode (test.test_package_version.PackageVersionTest)
Test that updating the changelog does not throw an exception on unicode characters.
----------------------------------------------------------------------
  File "c:\workspace\catkin_pkg\test\test_package_version.py", line 87, in test_update_changelog_unicode
    update_changelog_sections(missing_changelogs_but_forthcoming, '1.0.0')
  File "c:\workspace\catkin_pkg\src\catkin_pkg\package_version.py", line 160, in update_changelog_sections
    with open(changelog_path, 'wb') as f:
IOError: [Errno 13] Permission denied: 'c:\\users\\admin\\appdata\\local\\temp\\tmpvchrip'

### DIFF
--- a/test/test_package_version.py
+++ b/test/test_package_version.py
@@ -9,9 +9,9 @@ from catkin_pkg.package_version import bump_version
 from catkin_pkg.package_version import update_changelog_sections
 from catkin_pkg.package_version import update_versions
 
-from .util import in_temporary_directory
-
 import mock
+
+from .util import in_temporary_directory
 
 
 class PackageVersionTest(unittest.TestCase):
@@ -66,8 +66,7 @@ class PackageVersionTest(unittest.TestCase):
     @in_temporary_directory
     def test_update_changelog_unicode(self, directory=None):
         """Test that updating the changelog does not throw an exception on unicode characters."""
-        fd, temp_file = tempfile.mkstemp(dir=directory)
-        os.close(fd)
+        temp_file = os.path.join(directory, 'changelog')
         missing_changelogs_but_forthcoming = {}
         # Mock the Changelog object from catkin_pkg
         mock_changelog = mock.Mock()
--- a/test/test_package_version.py
+++ b/test/test_package_version.py
@@ -9,6 +9,8 @@ from catkin_pkg.package_version import bump_version
 from catkin_pkg.package_version import update_changelog_sections
 from catkin_pkg.package_version import update_versions
 
+from .util import temporary_file
+
 import mock
 
 
@@ -63,7 +65,7 @@ class PackageVersionTest(unittest.TestCase):
 
     def test_update_changelog_unicode(self):
         """Test that updating the changelog does not throw an exception on unicode characters."""
-        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+        with temporary_file() as temp_file:
             missing_changelogs_but_forthcoming = {}
             # Mock the Changelog object from catkin_pkg
             mock_changelog = mock.Mock()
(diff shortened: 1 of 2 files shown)


test/test_package_version.py
```diff
@@ -63,7 +63,7 @@ def test_update_versions(self):
 
     def test_update_changelog_unicode(self):
         """Test that updating the changelog does not throw an exception on unicode characters."""
-        with tempfile.NamedTemporaryFile() as temp_file:
+        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
```
dirk-thomas: With this chance the temporary file is not being deleted at all. That is not the desired behavior.

An alternative would be to create a temporary directory with the context manager instead - to ensure the directory is removed after the test as well as files in that directory can be read after being written.

seanyen-msft: Thank for the feedback! I changed it to use context manager.

test/util.py
```diff
@@ -34,3 +34,17 @@ def decorated(*args, **kwds):
...
+        pass
+
+    def __enter__(self):
+        self.temp_file = tempfile.NamedTemporaryFile(delete=False)
+        return self.temp_file
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if self.temp_file and os.path.exists(self.temp_file.name):
+            self.temp_file.close()
+            os.remove(self.temp_file.name)
```
dirk-thomas: I don't see how this class is significantly different to `tempfile.NamedTemporaryFile`?

seanyen-msft: Basically it drops the _O_TEMPORARY usage internally specified inside the NamedTemporaryFile when the delete=True. Under the hood, _open with _O_TEMPORARY is passing FILE_FLAG_DELETE_ON_CLOSE into Win32 API CreateFile and others cannot open the same file unless they are explicitly using FILE_SHARE_DELETE.
https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-createfilea

So my proposed workaround is to remove this unwanted effect since the code under test needs to touch the file but without knowing this _O_TEMPORARY (or FILE_SHARE_DELETE) details. Any thoughts on this?

dirk-thomas: These internals are not obvious from this patch and not even mentioned in the Python docs. They heavily rely on the implementation details (which could change anytime).

Hence I think my earlier proposal to create a `TemporaryDirectory` is the preferred approach.

seanyen-msft: That makes sense. I go use in_temporary_directory and mkstemp to replace the current implementation.

test/test_package_version.py
```diff
@@ -61,55 +63,57 @@ def test_update_versions(self):
...
-                                  'Forthcoming\n'
-                                  '-----------\n'
-                                  '* This is my changelog entry\n'
-                                  '* This is a line that has unicode' u'\xfc''\n'
-                                  '\n'
-                                  '0.0.9 (2017-01-30)\n'
-                                  '------------------\n'
-                                  '* This is old version.\n')
+        fd, temp_file = tempfile.mkstemp(dir=directory)
+        os.close(fd)
```
dirk-thomas: I don't think the `temp_file` needs to be created here. I is being written to by the called functions and later read back. A simple patch should be sufficient here:

```
temp_file = os.path.join(directory, 'changelog')
```

dirk-thomas:
Please also run the unit tests locally and address the linter warnings.

dirk-thomas:
Thank you for the patch and for iterating on it.

Issue #235, opened 2018-09-06
catkin_generate_changelog fails on hg/mercurial repositories


kmhallen:
Git repositories work as expected, but hg repositories fail and report tags that don't exist.

Commands to clone some hg repositories and reproduce:
```
cd /tmp && hg clone https://bitbucket.org/DataspeedInc/lusb
cd /tmp && hg clone https://bitbucket.org/DataspeedInc/dataspeed_can
cd /tmp && hg clone https://bitbucket.org/DataspeedInc/dbw_fca_ros
cd /tmp/lusb && catkin_generate_changelog
cd /tmp/dataspeed_can && catkin_generate_changelog
cd /tmp/dbw_fca_ros && catkin_generate_changelog
```

Output:
```
Found packages: lusb
Querying commit information since latest tag...
ERROR: The tag name '1' doesn't match the version pattern x.y.z

Found packages: dataspeed_can, dataspeed_can_msg_filters, dataspeed_can_msgs, dataspeed_can_tools, dataspeed_can_usb
Querying commit information since latest tag...
ERROR: The tag name '33' doesn't match the version pattern x.y.z

Found packages: dbw_fca, dbw_fca_can, dbw_fca_description, dbw_fca_joystick_demo, dbw_fca_msgs
Querying commit information since latest tag...
ERROR: The tag name '4' doesn't match the version pattern x.y.z
```

dirk-thomas:
Please consider to provide a pull request to add proper support for Mercurial repositories to `catkin_generate_changelog`.

PR #236, opened 2018-09-07
Fix catkin_generate_changelog for hg/mercurial repositories

Lohrer:
Fixes issue #235.


dirk-thomas:
The [man ](https://www.selenic.com/mercurial/hg.1.html)page of `hg` states:

> **latesttag:** List of strings. The global tags on the most recent globally tagged ancestor of this changeset.  If no such tags exist, the list consists of the single string "null".

Therefore I added the extra check for "null" in ae2692ad7d046cc1724848256afdce44548a95de.

PR #237, opened 2018-09-15
Exclude generated documentation files from flake8 test

cottsay:
For whatever reason, the `conf.py` generated in the `doc/_build` directory on Fedora doesn't have a trailing newline, so it fails the flake8 tests. Of course this isn't a problem if you don't build docs before running the tests. There really isn't any reason to run flake8 on anything generated in the docs anyway, so we can just skip those files.

### DIFF
--- a/test/test_flake8.py
+++ b/test/test_flake8.py
@@ -24,6 +24,8 @@ def test_flake8():
     # Configure flake8 using the .flake8 file in the root of this repository.
     style_guide = get_style_guide()
 
+    style_guide.options.exclude += ['*/doc/_build']
+
     stdout = sys.stdout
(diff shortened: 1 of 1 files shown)


cottsay:
Shoot, this works with flake8 `3.4.1` on Fedora, but it appears broken in Travis. I'll find another solution.

cottsay:
This solution seems to work on Ubuntu and Fedora. I have no clue why the constructor method didn't work - the docs are the same for `3.4.1` and `3.5.0`.

Note that this problem is also reproducible on Ubuntu if you generate docs before running tests.

dirk-thomas:
Thanks for the patch.

PR #238, opened 2018-10-04
warn not error (fail) for people wrongly using run_depend on version 2

k-okada:
Using run_depend with package.xml is not supported, but failing with error is too much for new users, we have so many question related this issue ->  see [Gogle result](https://www.google.co.jp/search?ei=JJG1W-GsMo_b8APs1J3gBg&q=the+manifest+%28with+format+version+2%29+must+not+contain+the+following+tags%3A+run_depend%22&oq=the+manifest+%28with+format+version+2%29+must+not+contain+the+following+tags%3A+run_depend%22 )

Note that we have not updated the translated tutorial which still assuming version 1, but actutal `catkin_create_pkg` has been updated for version  http://wiki.ros.org/ja/ROS/Tutorials/CreatingMsgAndSrv

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -605,6 +605,10 @@ def parse_package_string(data, filename=None, warnings=None):
             if not same_exec_depends:
                 pkg.exec_depends.append(deepcopy(dep))
         pkg.doc_depends = _get_dependencies(root, 'doc_depend')
+        for d in run_depends:
+            print('WARNING: The manifest of package "%s" (with format version %d) does not support <run_depend>%s</run_depend>. Please update your package.xml with <exec_depend>' % (pkg.name, pkg.package_format, d.name), file=sys.stderr)
+            pkg.build_export_depends.append(deepcopy(d))
+            pkg.exec_depends.append(deepcopy(d))
     pkg.test_depends = _get_dependencies(root, 'test_depend')
     pkg.conflicts = _get_dependencies(root, 'conflict')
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think this should be merged. The valid tags in a manifest file are specified in an REP. Allowing other tags and mapping them magically in the `catkin_pkg` implementation would break any other software trying to read the files which follow the specification strictly.

I would suggest to rather update the translated tutorials (or any other outdated documentation).

If you would like to have a more specific error message for some unknown tags (like `run_depend`) that sounds fine to me.

tfoote:
I agree that we shouldn't magically make an invalid package.xml work. It should be fixed. But the existing warnings are much more useful and could be left there as a helpful message to tell people what to fix to allow their package.xml to be parsed before quitting.

dirk-thomas:
@k-okada Are you planning to update the patch based on the feedback?

dirk-thomas:
@k-okada Friendly ping.

dirk-thomas:
Since there was no further response I guess just adding more detailed warnings isn't important enough. Therefore I will go ahead and close this. Please feel free to create a new PR just adding the more specific text to the error message.

k-okada:
@dirk-thomas sorry for late response, I look over the ping.

Coincidentally enough, I just asked "I tried to add message in my package and updated my package.xml and CMakelists.txt, after reading ros wiki, But somehow I can not run catkin build` question this morning, who just started ROS programming a few month ago. I have created new PR on #246 .

PR #239, opened 2018-10-05
Remove argparse Dependency

allenh1:
Since `argparse` is now a part of the language, runtime problems result from this (see ros/ros-overlay#687).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,19 @@
 #!/usr/bin/env python
 
 import os
+import sys
 
 from setuptools import setup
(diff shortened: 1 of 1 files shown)


allenh1:
Had a small linter issue, should br fixed now.

dirk-thomas:
Thanks for the patch. Same as https://github.com/ros-infrastructure/rospkg/commit/7d8aa9e9cc2d4cd3dbe7d8907bdabfd5a485869f.

PR #243, opened 2018-11-17
Test for metapackage with format 2

Nickolaim:
Produce different warning message for dependencies in format 1 (has to have `run_depends`) and formats 2 and 2 (`exec_depends`)


dirk-thomas:
Thank you for the improvements.

PR #245, opened 2019-01-03
evaluate condition for build_type export

### LINKED ISSUE #244: evaluate condition for build_type
As specified in ros-infrastructure/rep#171.

dirk-thomas:
Fixes #244.

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -148,7 +148,11 @@ class Package(object):
         :rtype: str
         :raises: :exc:`InvalidPackage`
         """
-        build_type_exports = [e.content for e in self.exports if e.tagname == 'build_type']
+        # for backward compatibility a build type without an evaluated
+        # condition is still being considered
+        build_type_exports = [
+            e.content for e in self.exports
+            if e.tagname == 'build_type' and e.evaluated_condition is not False]
         if not build_type_exports:
             return 'catkin'
         if len(build_type_exports) == 1:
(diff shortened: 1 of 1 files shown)


davidhodo:
Thanks.  I can confirm that I was able to build with both catkin and ament using the same package.xml after this fix.

src/catkin_pkg/package.py
```diff
@@ -148,7 +148,11 @@ def get_build_type(self):
         :rtype: str
         :raises: :exc:`InvalidPackage`
         """
-        build_type_exports = [e.content for e in self.exports if e.tagname == 'build_type']
+        # for backward compatibility a build type without an evaluated
+        # condition is still being considered
```
hidmic: nit: stating that `evaluated_condition` can only be `True`, `False` or `None` in the comment (or somewhere) may be helpful.

dirk-thomas: That is why I added the comment - to clarify that a not evaluated condition should be considered `True`. Since the default value is `None` and the value is optionally assigned to a boolean value I am not sure it is necessary to duplicate that in the comment. Please feel free to suggest to specific edit.

hidmic: Yeap, I realized, but not at first sight. Only after browsing the code. Maybe something like: _condition is still being considered **(i.e. evaluated_condition == None)**_ would do? In any case, it's pretty minor.

dirk-thomas: See aac58b3be261a9cc865ba52633113429adf118d6

PR #246, opened 2019-01-09
more detailed error message those who wrongly using run_depend on version 2

k-okada:
Added more detailed warnings when people uses `run_depend` on package.xml v2 or `exec_depend` on package.xml v1.
```
    xml, filename=filename, warnings=warnings)
  File "/home/k-okada/catkin_ws/ws_catkin_pkg/venv/lib/python2.7/site-packages/catkin_pkg-0.4.8-py2.7.egg/catkin_pkg/package.py", line 705, in parse_package_string
    raise InvalidPackage('Error(s):%s' % (''.join(['\n- %s' % e for e in errors])), filename)
catkin_pkg.package.InvalidPackage: Error(s) in package '/home/k-okada/catkin_ws/ws_catkin_pkg/src/test_pkg_2/package.xml':
Error(s):
- WARNING:
- WARNING: ACTION REQUIRED
- WARNING:
- WARNING: The manifest of package "test_pkg_2" (with format version 2) does not support <run_depend>. 
- WARNING: Please update your "test_pkg_2/package.xml" with <exec_depend>
- WARNING:
- WARNING:
```
c.f. #238

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ install_requires = [
 ]
 
 # argparse is part of the standard library since Python 2.7
-if sys.version_info[0] == 2 and sys.version_info[1] <= 7:
+if sys.version_info[0] == 2 and sys.version_info[1] < 7:
     install_requires.append('argparse')
 
 kwargs = {


src/catkin_pkg/package.py
```diff
@@ -696,7 +696,25 @@ def parse_package_string(data, filename=None, warnings=None):
         })
     nodes = [n for n in root.childNodes if n.nodeType == n.ELEMENT_NODE]
     unknown_tags = set([n.tagName for n in nodes if n.tagName not in known.keys()])
-    if unknown_tags:
+    if unknown_tags == set(['run_depend']) and pkg.package_format >= 2:
```
dirk-thomas: Instead of checking that the unknown tags are only containing `run_depend` I would suggest to print the message whenever `run_depend` is in `unknown_tags` (plus the existing format condition).

src/catkin_pkg/package.py
```diff
@@ -696,7 +696,25 @@ def parse_package_string(data, filename=None, warnings=None):
...
+        errors.append('\033[91mWARNING:')
```
dirk-thomas: Please do not use ASCII codes for colorizing the output here.

Please also avoid "empty" `WARNING:` lines.

src/catkin_pkg/package.py
```diff
@@ -696,7 +696,25 @@ def parse_package_string(data, filename=None, warnings=None):
...
+        errors.append('WARNING: ACTION REQUIRED')
```
dirk-thomas: Please also remove this line - since it is an error which will fail the process it is obvious that something is wrong.

src/catkin_pkg/package.py
```diff
@@ -696,7 +696,25 @@ def parse_package_string(data, filename=None, warnings=None):
...
+        errors.append('WARNING:')
+        errors.append('WARNING: The manifest of package "%s" (with format version %d) '
```
dirk-thomas: Please remove the `WARNING: ` prefix from the method since this is actual an error.

src/catkin_pkg/package.py
```diff
@@ -696,7 +696,25 @@ def parse_package_string(data, filename=None, warnings=None):
...
+                      'does not support <run_depend>. ' % (pkg.name, pkg.package_format))
+        errors.append('WARNING: Please update your "%s/package.xml" with <exec_depend>' % (pkg.name))
```
dirk-thomas: I would rephrase this sentence to:

> Replace <run_depend> tags with <exec_depend> tags

src/catkin_pkg/package.py
```diff
@@ -696,7 +696,25 @@ def parse_package_string(data, filename=None, warnings=None):
...
+        errors.append('WARNING:')
+        errors.append('WARNING:\033[0m')
+    elif unknown_tags == set(['exec_depend']) and pkg.package_format < 2:
+        errors.append('\033[91mWARNING:')
+        errors.append('WARNING: ACTION REQUIRED')
+        errors.append('WARNING:')
+        errors.append('WARNING: The manifest of package "%s" (with format version %d) '
+                      'does not support <exec_depend>. ' % (pkg.name, pkg.package_format))
+        errors.append('WARNING: Please update your "%s/package.xml" with <run_depend>' % (pkg.name))
```
dirk-thomas: Similar to the above but I would suggest to offer both possible paths forward:

> Either update to a newer format or replace <exec_depend> tags with <run_depend> tags.

src/catkin_pkg/package.py
```diff
@@ -696,7 +696,25 @@ def parse_package_string(data, filename=None, warnings=None):
         })
     nodes = [n for n in root.childNodes if n.nodeType == n.ELEMENT_NODE]
     unknown_tags = set([n.tagName for n in nodes if n.tagName not in known.keys()])
-    if unknown_tags:
+    if unknown_tags == set(['run_depend']) and pkg.package_format >= 2:
+        errors.append('\033[91mWARNING:')
+        errors.append('WARNING: ACTION REQUIRED')
+        errors.append('WARNING:')
+        errors.append('WARNING: The manifest of package "%s" (with format version %d) '
+                      'does not support <run_depend>. ' % (pkg.name, pkg.package_format))
```
dirk-thomas: Please remove the trailing space in the message. It is not necessary since each item is printed on a separate line anyway.

src/catkin_pkg/package.py
```diff
@@ -696,7 +696,25 @@ def parse_package_string(data, filename=None, warnings=None):
...
+    elif unknown_tags == set(['exec_depend']) and pkg.package_format < 2:
+        errors.append('\033[91mWARNING:')
+        errors.append('WARNING: ACTION REQUIRED')
+        errors.append('WARNING:')
+        errors.append('WARNING: The manifest of package "%s" (with format version %d) '
+                      'does not support <exec_depend>. ' % (pkg.name, pkg.package_format))
+        errors.append('WARNING: Please update your "%s/package.xml" with <run_depend>' % (pkg.name))
+        errors.append('WARNING:')
+        errors.append('WARNING:\033[0m')
+    elif unknown_tags:
```
dirk-thomas: I would suggest to always print this message first. And conditionally the more detailed messages afterwards providing specific suggestions for run/exec dependencies.

k-okada:
@dirk-thomas thanks for comments. I have fixed and committed newer version.

BTW, I still want to emphasize error message so that (specially) beginners do not confused with 'Traceback' and 'Error' message. How do you think raising 'InvalidPackage' exception with color formatted string.

```diff --git a/src/catkin_pkg/package.py b/src/catkin_pkg/package.py
index 9d49e61..d7f8269 100644
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -41,6 +41,7 @@ import sys
 import xml.dom.minidom as dom
 
 from catkin_pkg.condition import evaluate_condition
+from catkin_pkg.terminal_color import fmt
 
 PACKAGE_MANIFEST_FILENAME = 'package.xml'
 
@@ -714,7 +715,7 @@ def parse_package_string(data, filename=None, warnings=None):
                 errors.append('The "%s" tag must not contain the following children: %s' % (node.tagName, ', '.join([n.tagName for n in subnodes])))
 
     if errors:
-        raise InvalidPackage('Error(s):%s' % (''.join(['\n- %s' % e for e in errors])), filename)
+        raise InvalidPackage(fmt('@{rf}@{boldon}Error(s):%s' % (''.join(['\n- %s' % e for e in errors]))), filename)
```

src/catkin_pkg/package.py
```diff
@@ -698,6 +698,14 @@ def parse_package_string(data, filename=None, warnings=None):
     unknown_tags = set([n.tagName for n in nodes if n.tagName not in known.keys()])
     if unknown_tags:
         errors.append('The manifest (with format version %d) must not contain the following tags: %s' % (pkg.package_format, ', '.join(unknown_tags)))
+    if 'run_depend' in unknown_tags and pkg.package_format >= 2:
+        errors.append('The manifest of package "%s" (with format version %d) '
+                      'does not support <run_depend>.' % (pkg.name, pkg.package_format))
```
dirk-thomas: This message seems to almost be a duplicate of line 700. Maybe just add the package name to line 700 instead?

Same below.

src/catkin_pkg/package.py
```diff
@@ -698,6 +698,14 @@ def parse_package_string(data, filename=None, warnings=None):
...
+        errors.append('Please replace <run_depend> tags with <exec_depend> tags.')
+    elif 'exec_depend' in unknown_tags and pkg.package_format < 2:
+        errors.append('The manifest of package "%s" (with format version %d) '
+                      'does not support <exec_depend>.' % (pkg.name, pkg.package_format))
+        errors.append('Either update to a newer format or replace <exec_depend> tags.')
```
dirk-thomas: > ... replace <exec_depend> tags.

I would rather include the information what to replace it with: `... replace <exec_depend> tags with <run_depend> tags`

src/catkin_pkg/package.py
k-okada: That’s make sense

src/catkin_pkg/package.py
k-okada: Thanks. changed.

k-okada:
Sorry , i have closed this issue by force push origin/master to k-okada/master, moved PR to #248

PR #251, opened 2019-02-09
[Windows] Use multiprocessing.dummy on Windows for parallelism (plus adding Azure test pipeline.)

seanyen-msft:
This is an attempt to address the discussion in #250.

The change is to propose using multiprocessing.dummy (which is a threading wrapper) on Windows to avoid the restriction [`Safe importing of main module`](https://docs.python.org/2.7/library/multiprocessing.html#windows) (which requires the downstream tools to be updated to work). This way can keep parallelism but not requires the downstream tools to be modified.

Plus, this change adds an Azure pipeline yaml to exercise the test suite on Windows machines.

### DIFF
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -33,6 +33,11 @@
 """Library to find packages in the filesystem."""
 
 import os
+
+from .package import _get_package_xml
+from .package import PACKAGE_MANIFEST_FILENAME
+from .package import parse_package_string
+
 if os.name == 'nt':
     # https://docs.python.org/2/library/multiprocessing.html#windows
     #
@@ -44,10 +49,6 @@ if os.name == 'nt':
     import multiprocessing.dummy as multiprocessing
 else:
     import multiprocessing
-
-from .package import _get_package_xml
-from .package import PACKAGE_MANIFEST_FILENAME
-from .package import parse_package_string
 
 
 def find_package_paths(basepath, exclude_paths=None, exclude_subspaces=False):
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -32,8 +32,18 @@
 
 """Library to find packages in the filesystem."""
 
-import multiprocessing
 import os
+if os.name == 'nt':
+    # https://docs.python.org/2/library/multiprocessing.html#windows
+    #
+    # On Windows, using multiprocessing requires the scripts to be
+    # "Safe importing of main module" to avoid RuntimeError.
+    # This restriction requires downstream tools to be modified to
+    # work. Instead, let's fallback to threading wrapper to get
+    # around it and still keep the parallelism on Windows.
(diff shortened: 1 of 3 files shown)


seanyen-msft:
Here is the Azure CI result: https://dev.azure.com/seanyen2018/default/_build/results?buildId=386

test/test_packages.py
```diff
@@ -49,7 +49,7 @@ def test_find_packages_allowing_duplicates_with_no_packages():
 @in_temporary_directory
 def test_find_packages_invalid_version():
     version = ':{version}'
-    path = 'src/foo'
+    path = os.path.normpath('src/foo')
```
dirk-thomas: How does that work with the code three lines below: `path.split('/')`?

src/catkin_pkg/packages.py
```diff
@@ -32,13 +32,24 @@
...
 
+if os.name == 'nt':
+    # https://docs.python.org/2/library/multiprocessing.html#windows
+    #
+    # On Windows, using multiprocessing requires the scripts to be
+    # "Safe importing of main module" to avoid RuntimeError.
+    # This restriction requires downstream tools to be modified to
+    # work. Instead, let's fallback to threading wrapper to get
+    # around it and still keep the parallelism on Windows.
+    import multiprocessing.dummy as multiprocessing
```
dirk-thomas: What is the impact of using threading for this on Windows? Is the performance still better than without?

test/test_packages.py
dirk-thomas: ccc4e2889b49142e524448a3af2f92fe1074057c

src/catkin_pkg/packages.py
seanyen-msft: Well... ran some benchmark tests today and in a test case where it has 1000 packages (and each has dummy 1000 <build_depend> nodes), the result looks not good for threading wrapper in general on Windows...

On CPython 2.7:
- Disable parallel: 26 sec
- Using multiprocessing: 7.5 sec
- Using multiprocessing.dummy: 98.5 sec

On Anaconda Python 3.7:
- Disable parallel: 12.8 sec
- Using multiprocessing: 4.8 sec
- Using multiprocessing.dummy: 14.5 sec

(My environment is using `Intel® Xeon® Processor E5-1650 v4` 6-cores CPU.)
(My test case is `test_find_packages_with_large_amount_packages` in https://github.com/seanyen-msft/catkin_pkg/blob/windows_port_multiprocessing_disabled/test/test_packages.py)

dirk-thomas: The results don't look like this should be merged as is

seanyen-msft: That make senses. Since now we explored some approaches, any pointers what to investigate next at this moment?

seanyen-msft:
Closed it because it is very likely `multiprocessing.dummy` not going to be an acceptable solution.

PR #252, opened 2019-03-20
Fix parsing package.xml files with Unicode characters

zultron:
Unicode characters in package.xml break `parse_package_string()`.
Here's an example traceback caused by a package where the author's
name contained an umlaut:

    Traceback (most recent call last):
      [...]
      File "/usr/lib/python2.7/dist-packages/catkin_pkg/package.py", line 540, in parse_package_string
        raise InvalidPackage('The manifest contains invalid XML:\n%s' % ex, filename)
    catkin_pkg.package.InvalidPackage: The manifest contains invalid XML:
    'ascii' codec can't encode character u'\xf6' in position 256: ordinal not in range(128)

PR #253, opened 2019-03-22
support passing unicode to parse_package_string

dirk-thomas:
Redo of #252.

WIthout the patch the function `parse_package_string` works with:
* Python 2:
  * `str`
* Python 2:
  * `str`
  * `bytes`

With the patch it also works with `unicode` in Python 2.

The first commit adds the test coverage for the new use case (hence failing). The second commit implements the new support.


zultron:
This looks good to me.  Also replaces ros-infrastructure/rosdistro#133.  The unit test here is better, too.  Thanks!

PR #254, opened 2019-04-19
Make prerelease process work on Windows

rotu:
- Use shutil.which instead of manually finding git in path. This finds git.exe properly and is cleaner.
- Accept CR LF instead of just LF when trying to find version section

### DIFF
--- a/src/catkin_pkg/changelog_generator_vcs.py
+++ b/src/catkin_pkg/changelog_generator_vcs.py
@@ -65,6 +65,18 @@ class LogEntry(object):
         return False
 
 
+try:
+    from shutil import which
+except ImportError:
+    # fallback for python < 3.3
+    def which(cmd):
+        for path in os.getenv('PATH').split(os.path.pathsep):
+            file_path = os.path.join(path, cmd)
+            if os.path.isfile(file_path):
+                return file_path
+        return None
+
+
 class VcsClientBase(object):
 
     def __init__(self, path):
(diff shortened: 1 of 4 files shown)


src/catkin_pkg/changelog_generator_vcs.py
```diff
@@ -348,7 +352,6 @@ def get_vcs_client(base_path):
     client_types = [c.type for c in vcs_clients]
     if len(client_types) != len(set(client_types)):
         raise RuntimeError('Multiple vcs clients share the same type: %s' % ', '.join(sorted(client_types)))
-
```
dirk-thomas: Please revert this unrelated change.

src/catkin_pkg/cli/prepare_release.py
```diff
@@ -194,10 +195,16 @@ def push_changes(base_path, vcs_type, tag_name, dry_run=False):
 
 
 def _find_executable(vcs_type):
-    for path in os.getenv('PATH').split(os.path.pathsep):
-        file_path = os.path.join(path, vcs_type)
-        if os.path.isfile(file_path):
+    try:
+        file_path = shutil.which(vcs_type)
+        if file_path is not None:
             return file_path
```
dirk-thomas: Please move the last two lines out of the `try` / `except` block since they can't raise an `AttributeError`

src/catkin_pkg/changelog_generator_vcs.py
rotu: Oops. Done.

src/catkin_pkg/cli/prepare_release.py
rotu: I switched this over to a better approach. It's more obvious to catch an ImportError

dirk-thomas:
Thanks for the patch.

PR #257, opened 2019-07-12
Remember mismatched version's filename

paulbovbel:
Resolves `TypeError: 'dict_values' object does not support indexing` exception when using python3

### DIFF
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -172,9 +172,11 @@ def verify_equal_package_versions(packages):
     :raises: :exc:RuntimeError` If the version is not equal in all packages
     """
     version = None
+    filename = None
     for package in packages:
         if version is None:
             version = package.version
+            filename = package.filename
         elif package.version != version:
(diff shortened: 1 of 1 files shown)


src/catkin_pkg/packages.py
```diff
@@ -172,9 +172,11 @@ def verify_equal_package_versions(packages):
...
     """
     version = None
+    filename = None
     for package in packages:
         if version is None:
             version = package.version
+            filename = package.filename
         elif package.version != version:
-            raise RuntimeError('Two packages have different version numbers (%s != %s):\n- %s\n- %s' % (package.version, version, package.filename, packages[0].filename))
+            raise RuntimeError('Two packages have different version numbers (%s != %s):\n- %s\n- %s' % (package.version, version, package.filename, filename))
```
dirk-thomas: Please use `list(packages)[0]` instead of introducing the `filename` variable to keep the code more compact.

dirk-thomas:
Thanks for the fix.

PR #260, opened 2019-07-24
fixes overlaid develspace include dir order bug involving symlink

aurzenligl:
This bug occurs when `setup.bash` of overlaid workspace is sourced through symlink path rather than absolute path. This commit resolves symlinks in `order_path` algorithm, allowing to detect that: `ws1/devel/include` starts with `ws1/devel_symlink` if `ws1/devel_symlink` is a symlink to `ws1/devel`.

original PR with problem description:
https://github.com/ros/catkin/pull/1017

original error reproduction:
https://github.com/aurzenligl/study/tree/master/ros-includedirorder

### DIFF
--- a/src/catkin_pkg/workspaces.py
+++ b/src/catkin_pkg/workspaces.py
@@ -75,13 +75,6 @@ def order_paths(paths_to_order, prefix_paths):
     :param paths_to_order: list of paths
     :param prefix_paths: list of prefixes, must not end with '/'
     """
-    def contains(prefix, path):
-        prefix = os.path.realpath(prefix)
-        path = os.path.realpath(path)
-        return (path == prefix
-                or path.startswith(prefix + os.sep)
-                or (os.altsep and path.startswith(prefix + os.altsep)))
-
     # the ordered paths contains a list for each prefix plus one more which contains paths which do not match one of the prefix_paths
     ordered_paths = [[] for _ in range(len(prefix_paths) + 1)]
 
@@ -89,13 +82,19 @@ def order_paths(paths_to_order, prefix_paths):
         # put each directory into the slot where it matches the prefix, or last otherwise
         index = 0
         for prefix in prefix_paths:
-            if contains(prefix, path):
+            if _is_equal_or_in_parents(prefix, path):
                 break
             index += 1
         ordered_paths[index].append(path)
 
     # flatten list of lists
     return [j for i in ordered_paths for j in i]
+
+
+def _is_equal_or_in_parents(dir_, path):
+    dir_ = os.path.normcase(os.path.realpath(dir_))
+    path = os.path.normcase(os.path.realpath(path))
+    return path == dir_ or path.startswith(dir_ + os.sep)
 
 
 def ensure_workspace_marker(base_path):
--- a/src/catkin_pkg/workspaces.py
+++ b/src/catkin_pkg/workspaces.py
@@ -75,6 +75,13 @@ def order_paths(paths_to_order, prefix_paths):
     :param paths_to_order: list of paths
     :param prefix_paths: list of prefixes, must not end with '/'
     """
+    def contains(prefix, path):
+        prefix = os.path.realpath(prefix)
+        path = os.path.realpath(path)
+        return (path == prefix
+                or path.startswith(prefix + os.sep)
+                or (os.altsep and path.startswith(prefix + os.altsep)))
+
     # the ordered paths contains a list for each prefix plus one more which contains paths which do not match one of the prefix_paths
     ordered_paths = [[] for _ in range(len(prefix_paths) + 1)]
 
@@ -82,7 +89,7 @@ def order_paths(paths_to_order, prefix_paths):
         # put each directory into the slot where it matches the prefix, or last otherwise
         index = 0
         for prefix in prefix_paths:
-            if path == prefix or path.startswith(prefix + os.sep) or (os.altsep and path.startswith(prefix + os.altsep)):
+            if contains(prefix, path):
                 break
             index += 1
         ordered_paths[index].append(path)
--- a/test/test_workspaces.py
+++ b/test/test_workspaces.py
@@ -2,7 +2,6 @@ from __future__ import print_function
 
 import os
 import shutil
-import sys
 import tempfile
 import unittest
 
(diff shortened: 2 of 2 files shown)


test/test_workspaces.py
```diff
@@ -41,3 +42,18 @@ def test_order_paths(self):
         self.assertEqual(['foo', 'bar'], order_paths(['bar', 'foo'], ['foo']))
         self.assertEqual(['baz', 'foo', 'bar'], order_paths(['bar', 'foo', 'baz'], ['baz', 'foo']))
         self.assertEqual(['foo' + os.sep + 'bim', 'bar'], order_paths(['bar', 'foo' + os.sep + 'bim'], ['foo']))
+
+    @unittest.skipUnless(sys.platform.startswith('linux'), 'requires Unix')
```
dirk-thomas: Instead of checking the platform (e.g. it should also work on macOS) maybe check if the symlink can be created?

test/test_workspaces.py
```diff
@@ -41,3 +42,18 @@ def test_order_paths(self):
...
+    def test_order_paths_with_symlink(self):
+        try:
+            root_dir = tempfile.mkdtemp()
```
dirk-thomas: This line should be outside of the `try` block.

test/test_workspaces.py
```diff
@@ -41,3 +42,18 @@ def test_order_paths(self):
...
+            foo = os.path.join(root_dir, 'foo')
+            foo_inc = os.path.join(root_dir, 'foo/include')
```
dirk-thomas: Nitpick: use the variable `foo`: `foo_inc = os.path.join(foo, 'include')`

src/catkin_pkg/workspaces.py
```diff
@@ -75,14 +75,21 @@ def order_paths(paths_to_order, prefix_paths):
     :param paths_to_order: list of paths
     :param prefix_paths: list of prefixes, must not end with '/'
     """
+    def contains(prefix, path):
```
dirk-thomas: Please move this definition outside of the function and prefix it with an underscore. That way it can be accessed from the outside (e.g. from the test).

Also the order of the arguments seems unintuitive. I would expect the opposite order of arguments - reading: "path contains prefix".

src/catkin_pkg/workspaces.py
```diff
@@ -75,14 +75,21 @@ def order_paths(paths_to_order, prefix_paths):
...
+        prefix = os.path.realpath(prefix)
+        path = os.path.realpath(path)
```
dirk-thomas: I would suggest to use os.path.normcase(os.path.realpath(...))` for these two lines.

That simplifies the following condition to: `path == prefix or path.startswith(prefix + os.sep)`.

test/test_workspaces.py
aurzenligl: That's a good idea. Afaik `os.symlink` is simply not present when unavailable, let's use `hasattr` in skip condition.

test/test_workspaces.py
aurzenligl: I aimed at consistency - previous two testcases made tmpdir within try block. I put the line before the try block - in previous tests as well.

test/test_workspaces.py
aurzenligl: Valid remark, considering how tested algorithm avoids using naked '/' - done.

test/test_workspaces.py
```diff
@@ -41,3 +41,18 @@ def test_order_paths(self):
         self.assertEqual(['foo', 'bar'], order_paths(['bar', 'foo'], ['foo']))
         self.assertEqual(['baz', 'foo', 'bar'], order_paths(['bar', 'foo', 'baz'], ['baz', 'foo']))
         self.assertEqual(['foo' + os.sep + 'bim', 'bar'], order_paths(['bar', 'foo' + os.sep + 'bim'], ['foo']))
+
+    @unittest.skipUnless(hasattr(os, 'symlink'), 'requires symlink availability')
+    def test_order_paths_with_symlink(self):
+        root_dir = tempfile.mkdtemp()
+        try:
+            foo = os.path.join(root_dir, 'foo')
+            foo_inc = os.path.join(root_dir, os.path.join(foo, 'include'))
```
dirk-thomas: This change is incorrect. Please use my previous suggestion: https://github.com/ros-infrastructure/catkin_pkg/pull/260#discussion_r307357582

test/test_workspaces.py
dirk-thomas: Instead of checking for existence please check for an exception when invoking the function instead.

src/catkin_pkg/workspaces.py
aurzenligl: I guess `contains` meaning is ambiguous and depends on argument types:
- strings - `/etc/fstab` contains `/etc`
- paths (prefix understood as directory) - `/etc` contains `/etc/fstab`

It seems natural in Python to check containment by operator `in`, which uses container as rhs arg - e.g. `1 in [1]`.

I think `def _in_dir(path, dir_):` (then usage reads: `path in prefix, where prefix is dir`) would better explain its purpose, unless abbreviations in function names are frowned upon.

I noticed that private functions are usually placed immediately below functions which use them - I did the same.

src/catkin_pkg/workspaces.py
aurzenligl: Spot on, done.

aurzenligl:
@dirk-thomas, do you prefer review fixes visible in history as separate commits - or just squashed into original commits?

test/test_workspaces.py
aurzenligl: Obviously - corrected.

test/test_workspaces.py
aurzenligl: By calling function in skip condition expression I'd be creating a symlink, so I assume you recommend to forgo `skipUnless` in favor of silently passing the test by catching `AttributeError` when `symlink` attribute is not found in `os` module in test body?

src/catkin_pkg/workspaces.py
dirk-thomas: With Python 3 this becomes easier using `pathlib.Path.parents`. Maybe naming the function accordingly is even better: e.g. `_in_parents` which would map directly to what Python 3 code would express?

test/test_workspaces.py
aurzenligl: I settled for:
```
            try:
                os.symlink(foo, foo_ln)
            except (AttributeError, OSError):
                self.skipTest('requires symlink availability')
```

src/catkin_pkg/workspaces.py
aurzenligl: I thought about that (or name mentioning "ancestor"), but both terms don't involve checking for identity. I.e. `/etc` is not among parents/ancestors of `/etc`
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parents

dirk-thomas: Since it is only an internal helper function maybe just very verbose: `_is_equal_or_parent`?

aurzenligl: I agree with compound name. However, `/etc` is not a parent of `/etc/apt/sources.list`, and I'd prefer to have something resembling container on rhs. Which brings us to: `_is_equal_or_descendant`. Does it sound ok?

dirk-thomas: `_is_equal_or_in_parents`? Since in Python 3 would do: `p1 == p2 or p1 in p2.parents`

aurzenligl: Reasonable, I've done that.

test/test_workspaces.py
```diff
@@ -25,9 +25,9 @@ def test_ensure_workspace_marker(self):
             shutil.rmtree(root_dir)
 
     def test_get_spaces(self):
-        self.assertEqual([], get_spaces([]))
+        root_dir = tempfile.mkdtemp()
         try:
-            root_dir = tempfile.mkdtemp()
+            self.assertEqual([], get_spaces([]))
```
dirk-thomas: Please move this line back outside where it was since it can't raise and doesn't need the `root_dir`.

aurzenligl: Done.

dirk-thomas:
LGTM. Thanks for addressing this and iterating on the patch.

> do you prefer review fixes visible in history as separate commits - or just as amends into original commits?

In general it is easier to review incremental commits since it is quicker to check what is changed in each iteration without re-reviewing the full patch. Luckily this one is fairly small - so no problem.

PR #264, opened 2019-08-22
remove trusty, utopic, vivid, wily from list of suites

dirk-thomas:
Making the latest not-EOL LTS - Xenial - the lowest watermark.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -11,7 +11,7 @@ Depends3: python3-catkin-pkg-modules (>= 0.4.13), python3-dateutil, python3-docu
 Conflicts: catkin, python3-catkin-pkg
 Conflicts3: catkin, python-catkin-pkg
 Copyright-File: LICENSE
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Should we drop Jessie while we're pruning as it's also out of support?

dirk-thomas:
Jessie is still a target platform of an active ROS distro - Kinetic. The REP doesn't state that we will retire Jessie support ahead of time of the EOL date of Kinetic. Therefore I would just keep it in the list of suites until then.

nuclearsandwich:
> Jessie is still a target platform of an active ROS distro - Kinetic. The REP doesn't state that we will retire Jessie support ahead of time of the EOL date of Kinetic. Therefore I would just keep it in the list of suites until then.

We've stopped building new debs for Jessie, and in fact there is nothing in the ROS repos for Jessie except the bootstrap packages since we didn't rebuild Jessie after the security event. But I'm not too fussed about keeping it if others find that a good reason for it to stay.

PR #265, opened 2019-09-25
Prevent Parallization on Windows

ooeygui:
Windows has an environment size limitation, which catkin can overflow during nested operations.
This fixes: https://github.com/ms-iot/ROSOnWindows/issues/148


dirk-thomas:
What has changed since #250 which proposed the same change has been closed?

ooeygui:
@dirk-thomas We're working to unblock customer deployments, and a customer is hitting this issue.

PR #268, opened 2019-10-11
check if ros xml schema link is present

fabolhak:
Adds a function to check whether a ROS schema is referenced in the package.xml.

As discussed in https://github.com/colcon/colcon-ros/pull/81.

`nosetests` run without problems. However, didn't check `nosetests3` (don't know how to install).

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -503,7 +503,7 @@ def _get_package_xml(path):
 
 def check_xml_schema_link(path):
     """
-    Checks whether a link to an ROS package schema exists for the given package manifest.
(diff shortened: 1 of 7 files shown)


fabolhak:
wow, this linter is really tough

dirk-thomas:
Thank you for the contribution.

I made some minor modifications - I hope they look good to you:
* most important provide two functions with a slightly different name: `has_ros_schema_reference` and `has_ros_schema_reference_string`
* ensure the right encoding in Python 2 before trying to parse the data (same as in another existing function)
* catch exception in case the XML data is invalid (same as in another existing function)
* simplified the tests to just pass strings and avoid repetition

PR #269, opened 2019-11-12
Add a __repr__ for Dependency

Rayman:
During debuging, I've made a simple `__repr__` for Dependency. It helped
me to visualize what was going on.

It prints dependencies like this:
```py
Dependency(name='foo', version_lt='3.0.0', condition='$ROS_VERSION == 2')
```

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -346,6 +346,14 @@ class Dependency(object):
     def __str__(self):
         return self.name
 
+    def __repr__(self):
+        kv = []
+        for slot in self.__slots__:
+            attr = getattr(self, slot, None)
+            if attr:
+                kv.append('{}={!r}'.format(slot, attr))
+        return '{}({})'.format(self.__class__.__name__, ', '.join(kv))
+
     def evaluate_condition(self, context):
         """
         Evaluate the condition.
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -350,7 +350,7 @@ class Dependency(object):
         kv = []
         for slot in self.__slots__:
             attr = getattr(self, slot, None)
-            if attr:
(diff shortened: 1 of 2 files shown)


src/catkin_pkg/package.py
```diff
@@ -346,6 +346,14 @@ def __hash__(self):
     def __str__(self):
         return self.name
 
+    def __repr__(self):
+        kv = []
+        for slot in self.__slots__:
+            attr = getattr(self, slot, None)
+            if attr:
```
dirk-thomas: Should this condition be changed to `attr is not None`? I would guess that it would be valuable to see when e.g. `evaluated_condition` has a value of `False`.

Rayman: You are right, that is valuable. I've modified the code and added a test.

test/test_package.py
```diff
@@ -118,6 +118,16 @@ def test_init_dependency(self):
         dep = Dependency('foo', condition='foo <= bar or bar >= baz')
         self.assertFalse(dep.evaluate_condition({}))
 
+    def test_dependency_repr(self):
+        dep = Dependency('foo', condition='$foo == 2')
+        assert repr(dep) == "Dependency(name='foo', condition='$foo == 2')"
+        self.assertTrue(dep.evaluate_condition({'foo': 2}))
+        assert repr(dep) == "Dependency(name='foo', condition='$foo == 2', evaluated_condition=True)"
+
+        dep = Dependency('foo', condition='$foo == 2')
```
dirk-thomas: Nitpick: is this line really necessary? Can't it instead use the already existing `dep`?

Rayman: You are right, you can re-evaluate the dependency condition. I've updated the PR

dirk-thomas:
Thanks for the improvement!

PR #270, opened 2019-11-18
fix some escape sequences

tgpfeiffer:
Escaping in regexes has not been done correctly in a number of places, which recent Python versions warn about.

### DIFF
--- a/src/catkin_pkg/changelog_generator_vcs.py
+++ b/src/catkin_pkg/changelog_generator_vcs.py
@@ -155,7 +155,7 @@ class GitClient(VcsClientBase):
         if result_tag['returncode']:
             raise RuntimeError('Could not fetch tags:\n%s' % result_tag['output'])
         # Parse a comma-separated list of refname decorators out of the log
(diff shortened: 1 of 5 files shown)


src/catkin_pkg/changelog_generator_vcs.py
```diff
@@ -240,7 +240,7 @@ def _determine_repo_hosting(self):
                 break
 
     def _replace_github_issue_references(self, line):
-        valid_name = '[\\w\._-]+'
+        valid_name = '[\\w._-]+'
```
tgpfeiffer: The dot (`.`) doesn't need escaping when it's in a character span.

src/catkin_pkg/cmake.py
```diff
@@ -77,4 +77,4 @@ def configure_string(template, environment):
     def substitute(match):
         var = match.group(0)[1:-1]
         return environment[var]
-    return re.sub('\@[a-zA-Z0-9_]+\@', substitute, template)
+    return re.sub('@[a-zA-Z0-9_]+@', substitute, template)
```
tgpfeiffer: Maybe I don't understand what's going here, but AFAIK the `@` character doesn't need any escaping, so I removed the backslash.

dirk-thomas:
Thanks for the patch.

PR #271, opened 2019-12-03
Add a Suite3 option with Ubuntu Focal.

nuclearsandwich:
This configuration option requires two recent changes in stdeb and
ros_release_python:
* https://github.com/astraw/stdeb/pull/147 for Suite3 support from stdeb (not yet merged or released)
* https://github.com/ros-infrastructure/ros_release_python/pull/33 for `--include` support on the current bootstrap repository.

The option allows a different set of suites to be set for Python 3
releases allowing us to forestall releasing Python2 versions of these
packages for platforms where we don't plan to support Python 2.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -12,6 +12,7 @@ Conflicts: catkin, python3-catkin-pkg
 Conflicts3: catkin, python-catkin-pkg
 Copyright-File: LICENSE
 Suite: xenial yakkety zesty artful bionic cosmic disco eoan jessie stretch buster
+Suite3: xenial yakkety zesty artful bionic cosmic disco eoan focal jessie stretch buster
 X-Python3-Version: >= 3.4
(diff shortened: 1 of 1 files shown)


dirk-thomas:
ros-infrastructure/ros_release_python#33 has nothing to do with this change.

nuclearsandwich:
> [ros-infrastructure/ros_release_python#33](https://github.com/ros-infrastructure/ros_release_python/pull/33) has nothing to do with this change.

`--include` doesn't work at all, suite3 or no without that change. This PR won't have the intended effect without it.

dirk-thomas:
> `--include` doesn't work at all, suite3 or no without that change. This PR won't have the intended effect without it.

Yes, `--include` was broken. Using `Suite3` or just `Suite` doesn't make any difference. Therefore these two changes are orthogonal.

PR #276, opened 2020-02-02
Changes 'install' to 'catkin_install_python' in the CMakeLists.txt te…

swinterbotix:
…mplate when installing Python Executable Scripts


dirk-thomas:
Thanks for the improvement.

PR #279, opened 2020-02-23
[enhancement]evaluating more than 1 condition fixes #277

### LINKED ISSUE #277: Logical operators require parentheses around operands
Once we get to three or more logical operators, the evaluation fails due to an assertion here:

https://github.com/ros-infrastructure/catkin_pkg/blob/27013d87262bef0d6ac7b9642aa4ed18d9ce21cd/src/catkin_pkg/condition.py#L61

Example input:

```xml
<depend condition="$FOO != ON and $FOO != 1 and $FOO != true">
```

If we add parentheses, then the expression evaluates correctly:

```xml
<depend condition="($FOO != ON and $FOO != 1) and $FOO != true">
```

I would think that the expression without parenthesis should work fine (as it would in a programming language).

nepython:
This is a improvement which fixes issue #277(A feature request to be able to apply more than 1 condition in a go).

### DIFF
--- a/src/catkin_pkg/condition.py
+++ b/src/catkin_pkg/condition.py
@@ -57,8 +57,8 @@ def _evaluate(parse_results, context):
         # return literal value
         return parse_results
 
-    assert len(parse_results) >= 3 and len(parse_results) % 2 == 1
     # recursion
+    assert len(parse_results) >= 3 and len(parse_results) % 2 == 1
 
     # handle logical operators
     if parse_results[1] == 'and':
--- a/src/catkin_pkg/condition.py
+++ b/src/catkin_pkg/condition.py
@@ -71,6 +71,7 @@ def _evaluate(parse_results, context):
                       else parse_results[2:], context)
 
     # handle comparison operators
+    assert len(parse_results) == 3
     operators = {
         '==': operator.eq,
         '!=': operator.ne,
--- a/src/catkin_pkg/condition.py
+++ b/src/catkin_pkg/condition.py
@@ -57,16 +57,18 @@ def _evaluate(parse_results, context):
         # return literal value
         return parse_results
 
+    assert len(parse_results) >= 3 and len(parse_results) % 2 == 1
     # recursion
-    assert len(parse_results) == 3
 
     # handle logical operators
     if parse_results[1] == 'and':
         return _evaluate(parse_results[0], context) and \
-            _evaluate(parse_results[2], context)
+            _evaluate(parse_results[2] if len(parse_results) == 3
+                      else parse_results[2:], context)
     if parse_results[1] == 'or':
         return _evaluate(parse_results[0], context) or \
-            _evaluate(parse_results[2], context)
+            _evaluate(parse_results[2] if len(parse_results) == 3
+                      else parse_results[2:], context)
 
     # handle comparison operators
     operators = {
--- a/test/test_package.py
+++ b/test/test_package.py
@@ -118,6 +118,31 @@ class PackageTest(unittest.TestCase):
         dep = Dependency('foo', condition='foo <= bar or bar >= baz')
         self.assertFalse(dep.evaluate_condition({}))
 
+        # Testing for more than 1 conditions
+        dep = Dependency('foo', condition='foo > bar and bar < baz and foo > bar')
+        self.assertTrue(dep.evaluate_condition({}))
+
+        dep = Dependency('foo', condition='foo <= bar and bar >= baz and foo > bar')
+        self.assertFalse(dep.evaluate_condition({}))
+
+        dep = Dependency('foo', condition='foo > bar or bar < baz or foo <= bar')
+        self.assertTrue(dep.evaluate_condition({}))
+
+        dep = Dependency('foo', condition='foo <= bar or bar >= baz or foo <= bar')
+        self.assertFalse(dep.evaluate_condition({}))
+
+        dep = Dependency('foo', condition='foo <= bar and bar < baz or foo > bar')
+        self.assertTrue(dep.evaluate_condition({}))
+
+        dep = Dependency('foo', condition='foo <= bar or bar < baz and foo < bar')
+        self.assertFalse(dep.evaluate_condition({}))
+
+        dep = Dependency('foo', condition='foo > bar and bar >= baz or foo > bar')
+        self.assertTrue(dep.evaluate_condition({}))
+
+        dep = Dependency('foo', condition='foo <= bar or bar >= baz and foo < bar')
+        self.assertFalse(dep.evaluate_condition({}))
+
     def test_dependency_repr(self):
         dep = Dependency('foo', condition='$foo == 2')
         assert repr(dep) == "Dependency(name='foo', condition='$foo == 2')"


nepython:
@dirk-thomas , can you please review the changes.

src/catkin_pkg/condition.py
```diff
@@ -58,6 +58,20 @@ def _evaluate(parse_results, context):
         return parse_results
 
     # recursion
+    if len(parse_results) != 3:
```
dirk-thomas: The following code only works for specific `len(oarse_results)`, e.g. `2` wouldn't work. Please add an assert to only handle those explicitly.

src/catkin_pkg/condition.py
```diff
@@ -58,6 +58,20 @@ def _evaluate(parse_results, context):
...
+        arg = parse_results[0]
+        condition_len = len(parse_results)
```
dirk-thomas: Nitpick: please avoid single-use variables.

src/catkin_pkg/condition.py
```diff
@@ -58,6 +58,20 @@ def _evaluate(parse_results, context):
...
+        parse_list = []
+
+        # After the loop ends, a final evaluation is done
+        for i in range(int(condition_len/2-1)):
```
dirk-thomas: Nitpick: please move the `- 1` out of the `int()` call.

(Might be obsolete with the next comment.)

src/catkin_pkg/condition.py
```diff
@@ -58,6 +58,20 @@ def _evaluate(parse_results, context):
...
+            parse_list = [arg]
+            parse_list.extends(parse_results[2*i+1:2*i+3])
```
dirk-thomas: Since you use `i` always with a factor of `2` maybe using `step=2` in the `range()` function would be better.

nepython:
I will make the recommended changes.

> Please also add unit tests for the newly supported cases as well as make sure that all tests pass.
> 

Regarding that,
The tests failed as I have exceeded the number of maximum commits in pull request allowed by flake.
I think, I won't be able to squash my commits as I had not cloned them priorily. If you check the commit prior to last in this pr, it passed the tests. So a request, please squash and merge them after I implement your recommendations.

dirk-thomas:
> The tests failed as I have exceeded the number of maximum commits in pull request allowed by flake.

I don't think such a limitation exists.

Please re-read the error message from the CI build (trailing whitespace) and commit a fix for it to the branch of your PR.

> I won't be able to squash my commits as I had not cloned them priorily.

You can easily clone your forked repo, switch to the branch of this PR, and apply any kind of change necessary - if need be even a squash of commit (which I highly doubt is necessary).

> please squash and merge them after I implement your recommendations.

The PR will **not** be merged until it passes CI. Please address the above comments and CI will pass.

src/catkin_pkg/condition.py
```diff
@@ -58,6 +58,20 @@ def _evaluate(parse_results, context):
...
+            arg = _evaluate(parse_list, context)
+
```
dirk-thomas: This line contains trailing white spaces which is flagged by the linter.

nepython:
@dirk-thomas , you were correct. 

> This line contains trailing white spaces which is flagged by the linter.
>
Thanks for pointing it out. :relaxed:
I have applied all your suggested changes, can you please review them.

src/catkin_pkg/condition.py
dirk-thomas: It it really the case that `parse_results` can have any number: `4`, `5`, `6`, ...?

(If yes, the following `if` can be removed since the `assert` ensures it is always `True`.)

src/catkin_pkg/condition.py
dirk-thomas: Still pending.

dirk-thomas:
> Please also add unit tests for the newly supported cases

Still pending.

src/catkin_pkg/condition.py
nepython: `len(parse_results)` is supposed to be a odd number, So, I have put that condition in `if`.

src/catkin_pkg/condition.py
nepython: Done

src/catkin_pkg/condition.py
```diff
@@ -57,8 +57,27 @@ def _evaluate(parse_results, context):
         # return literal value
         return parse_results
 
+    assert len(parse_results) >= 3
+
     # recursion
-    assert len(parse_results) == 3
+
+    if (len(parse_results) % 2) != 0:
```
dirk-thomas: What happens in the `else` case? The following code requires the `parse_results` to have three items which might not be the case anymore.

src/catkin_pkg/condition.py
```diff
@@ -57,8 +57,27 @@ def _evaluate(parse_results, context):
...
+        parse_list = parse_results[0:3]
+
+        # After the loop ends, a final evaluation is done
+        for i in range(1, int(len(parse_results))-2, 2):
```
dirk-thomas: Nitpick: no need for `int()` anymore.

src/catkin_pkg/condition.py
nepython: https://github.com/nepython/catkin_pkg/blob/033ab12454d88f33f7f5a68cee7e99fa6ac6c3a6/src/catkin_pkg/condition.py#L60-L64
This code requires that there be odd number of arguments in parse result greater than 3.
eg. 3, 5, 7,..

nepython: These code snippet is supposed to allow only odd number of arguments greater than 3 for parse_results 
eg. 3,5,7,.. as this is expected.

dirk-thomas: I understand what the condition does. But if the length is e.g. 4 it doesn't go into the `if` block. But the code below the if block is expecting exactly 3 and won't work correctly with a length of 4, 6, 8, etc.

nepython: The parse_results should look like:-
If one `or` / `and` is present `[['foo', '<=', 'bar'], 'or', ['bar', '>=', 'baz']] `args = **3**
If two `or` / `and` are present `[['foo', '<=', 'bar'], 'and', ['bar', '<', 'baz'], 'or', ['foo', '>', 'bar']]` args = **5**
and so on...
So, I don't think 4, 6, 8, etc. are even possible.

dirk-thomas: Then please readd an assertion which ensures that this precondition is met.

src/catkin_pkg/condition.py
```diff
@@ -57,8 +57,27 @@ def _evaluate(parse_results, context):
         # return literal value
         return parse_results
 
+    assert len(parse_results) >= 3
```
nepython: This is ensuring the same.:blush:

nepython: I thought that the removed assertion was adding to redundancy in the code, so removed it.

dirk-thomas: No, this line does **not** ensure that. It only ensures >= 3 but not that it must be odd.

nepython: Done

nepython: @dirk-thomas , can you please review it?

nepython:
>###### @dirk-thomas, can you please review it? 
>
@dirk-thomas, Gentle Reminder :smile:

nepython:
>###### @dirk-thomas, can you please review it? 
>
@dirk-thomas, Gentle Reminder :smile:

nepython:
@dirk-thomas Gentle Reminder :smile:

nepython:
@dirk-thomas, can you please review the code. It's been almost a week.

dirk-thomas:
@nepython I will review the pull request as soon as I find time for it. Poking me four times won't generate more time on my side unfortunately.

dirk-thomas:
I took another look at the patch and the logic with this loop and additional recursion doesn't make any sense to me. The function already does check for the logical operators `and` and `or` below and does invoke itself recursively. I don't see why all of that should be duplicated.

It should be completely sufficient to reduce this patch to two simple changes:

* change the existing assert to `assert len(parse_results) >= 3 and len(parse_results) % 2 == 1` to allow parser results with odd number of items greater 3 and
* modify the second argument for the recursive invocation to `parse_results[2] if len(parse_results) == 3 else parse_results[2:]`.

nepython:
@dirk-thomas, thanks for the review :smiley:
> The function already does check for the logical operators and and or below and does invoke itself recursively. I don't see why all of that should be duplicated.

```
                if (parse_results[i + 2] == 'and'):
                    return False
                if (parse_results[i + 2] == 'or'):
                    parse_list = parse_results[i + 1:i + 4]
```
I had implemented these to account for short-circuiting which is quite useful when evaluating logical expressions but if you feel it is not useful, I will remove the same and do it as per your suggestion.

dirk-thomas:
> I had implemented these to account for short-circuiting which is quite useful when evaluating logical expressions.

What do you mean with this? Why is it being useful? Please try to be more detailed / precise in your comment.

> after applying your second suggested change(which I assume makes the loop go on until all conditions have been evaluated or ...

My suggestion is for the while diff to be reduced to my previous comment. There shouldn't be any need for a loop in the first place.

nepython:
@dirk-thomas, thanks for the quick response
>What do you mean with this? Why is it being useful? Please try to be more detailed / precise in your comment.
 
I think I have mentioned it in the code snippet https://github.com/ros-infrastructure/catkin_pkg/pull/279#issuecomment-596962808 next to every line how it is helping in the short circuiting by stopping evaluation of further conditions in cases when we know that at the end result would be `True`(for `or`)/`False`(for `and`), I am mentioning the same below too.

> There shouldn't be any need for a loop in the first place.

Is your suggestion to implement something like:-
```
if len(parse_results>3):
  if _evaluate(parse_results[0:3],context): #first condition evaluates to True
    if parse_results[3]=='or': #If it's a 'or' and condition evaluates to True, further evaluations can be skipped
      return True 
    #Recursion
    else:
      parse_results = parse_results[2:]
      _evaluate(parse_results,context)
  elif parse_results[3]=='or':  #If it's a 'or' and condition evaluates to False, still futher evaluation needs to be done
    parse_results = parse_results[2:]
    _evaluate(parse_results,context)
  else: #It's an 'and' and entire condition evaluates to False in that case
    return False
#Normal single condition evaluation 
```

dirk-thomas:
> Is your suggestion to implement something like:

No, my suggestion is literally in the above comment: https://github.com/ros-infrastructure/catkin_pkg/pull/279#issuecomment-596875191 The code snippets in that comment are the complete diff I think is needed.

nepython:
>`parse_results[2] if len(parse_results) == 3 else parse_results[2:]`

@dirk-thomas, I do get what you mean but can you please consider edge cases such as one below:-
Suppose the condition is `foo==bar and bar==bar or foo==foo` (say` foo`!=`bar`) will lead to `False`, whereas in python it evaluates to `True`, hence I had put those extra checks to ensure that is also the case here.

Also, short-circuiting will help reduce the total recursion calls as it would simply return a value as I have explained above and not evalute all the conditions and then take decision, should definitely help.

Rest is your call, I can definitely make those changes just thought that few cases as the one I have mentioned above were amiss.

dirk-thomas:
> please consider edge cases such as one below

Then please add a unit test for that edge case ensuring that it does return the expected result.

nepython:
Sure, does everything else in https://github.com/ros-infrastructure/catkin_pkg/pull/279#issuecomment-597167450 seems fine? I will squash all my commits and send one with these lines them finally.

dirk-thomas:
> does everything else in #279 (comment) seems fine?

No, as mentioned multiple time there is no need to duplicate all this logic. If you really want to limit recursion then please do so in a separate PR. This one is to support multiple conditions.

dirk-thomas:
@nepython Friendly ping.

nepython:
@dirk-thomas, extremely sorry got a little busy. I have sent the patch as per your suggestions in [comment](https://github.com/ros-infrastructure/catkin_pkg/pull/279#issuecomment-596875191). Please have a look. Shall I open another PR for limiting the number of recursions, this will help if there happen to be many conditions involved. :)

src/catkin_pkg/condition.py
```diff
@@ -57,8 +57,13 @@ def _evaluate(parse_results, context):
...
         return parse_results
 
+    assert len(parse_results) >= 3 and len(parse_results) % 2 == 1
     # recursion
-    assert len(parse_results) == 3
+
+    if len(parse_results) == 3:
+        parse_results = parse_results[:3]
+    else:
+        _evaluate(parse_results[2:], context)
```
dirk-thomas: I don't think we need another case were `_evaluate()` is being called recursively. Instead the existing calls can be updated as described in https://github.com/ros-infrastructure/catkin_pkg/pull/279#issuecomment-596875191 to pass the right second argument.

dirk-thomas:
> Shall I open another PR for limiting the number of recursions, this will help if there happen to be many conditions involved.

The recursion limit is 3000 (at least in Python 3.8) which should be plenty and I don't expect we ever have the case of that many conditions. But I will leave that up to you to decide - if you think it is valuable please open a new PR once this one has been merged.

src/catkin_pkg/condition.py
nepython: @dirk-thomas, I am unsure I got you correctly can you please specifically point which *existing calls* you mentioned to be updated? :smiley:

dirk-thomas: https://github.com/ros-infrastructure/catkin_pkg/blob/dcd329d5291fc2bbe4b481f59f7eca7d9660cd75/src/catkin_pkg/condition.py#L64-L69

nepython:
@dirk-thomas, I think this is finally done on my side. Not sure why is **Travis** not running at all, might be down for maintenance. I have locally run `nosetests -s --tests test` and everything is passing.

> The recursion limit is 3000 (at least in Python 3.8) which should be plenty and I don't expect we ever have the case of that many conditions.

I think then there is no need for it then.
If you think there are still improvements please suggest I would be happy to make the necessary changes. :smile:

dirk-thomas:
> Not sure why is Travis not running at all, might be down for maintenance.

You might want to check https://www.traviscistatus.com/ for that information rather than just guessing.

Anyway I committed to tiny changes (see ae9fff5d2dfdc771416054b395250faf3fe618f6 and 657e0ec790b9faa53dfa50c9540ced24b85a603d) and Travis is processing the latest commit now.

dirk-thomas:
Thanks for the improvement and for iterating on the ticket.

PR #280, opened 2020-03-24
Speed up topological_order by using a per-package cache of all run_depends

seeraven:
When using a large set of packages (e.g., more than 1000), catkin spends a lot of time calculating all the dependencies of all found packages. In my case, it was about 26 seconds. By caching the dependencies of the run_depends, it is now down to 4 seconds.

### DIFF
--- a/src/catkin_pkg/topological_order.py
+++ b/src/catkin_pkg/topological_order.py
@@ -50,6 +50,7 @@ class _PackageDecorator(object):
         self.message_generator = message_generators[0] if message_generators else None
         # full includes direct build depends and recursive run_depends of these build_depends
         self.depends_for_topological_order = None
+        self._recursive_run_depends_for_topological_order = None
 
     def __getattr__(self, name):
         if name.startswith('__'):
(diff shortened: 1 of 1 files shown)


src/catkin_pkg/topological_order.py
```diff
@@ -50,6 +50,7 @@ def __init__(self, package, path):
         self.message_generator = message_generators[0] if message_generators else None
         # full includes direct build depends and recursive run_depends of these build_depends
         self.depends_for_topological_order = None
+        self.run_depends_for_topological_order = None
```
dirk-thomas: This member should be renamed to `_recursive_run_depends_for_topological_order` to a) be private and b) contain "recursive" as part of the name.

src/catkin_pkg/topological_order.py
```diff
@@ -85,7 +86,22 @@ def calculate_depends_for_topological_order(self, packages):
...
+
+        This method is the cached variant of _add_recursive_run_depends().
+        Only packages which are in the passed packages are added and recursed into.
+
+        :param packages: dict of name to ``_PackageDecorator``
+        :param depends_for_topological_order: set to be extended
+        """
+        if self.run_depends_for_topological_order is None:
+            self.run_depends_for_topological_order = set()
+            self._add_recursive_run_depends(packages, self.run_depends_for_topological_order)
```
dirk-thomas: Atm it seems that `run_depends_for_topological_order` contains its own package name which I don't think is desired (since a package is not allowed to depend on itself).

src/catkin_pkg/topological_order.py
```diff
@@ -85,7 +86,22 @@ def calculate_depends_for_topological_order(self, packages):
                 # avoid function call to improve performance
                 # check within the loop since the set changes every cycle
                 continue
-            packages[name]._add_recursive_run_depends(packages, self.depends_for_topological_order)
+            packages[name]._add_recursive_run_depends_cached(packages, self.depends_for_topological_order)
+
+    def _add_recursive_run_depends_cached(self, packages, depends_for_topological_order):
```
dirk-thomas: Instead of introducing a new method how about updating `_add_recursive_run_depends` instead to do the caching directly?

seeraven:
> Please also include reproducible steps how you tested / benchmarked the change.

I've created the repository https://github.com/seeraven/catkin_pkg_benchmark that contains an anonymous version of my workspace and simulates the steps taken by catkin. Actually catkin calls the `topological_order_packages` function multiple times, but the benefit of caching can be already seen when calling the function only once.
On my workstation, the timings are:
```
Original catkin_pkg:
Run 1: Found 1460 packages in 0.4 s
Run 1: Topological order took 7.5 s
Run 2: Found 1460 packages in 0.4 s
Run 2: Topological order took 7.4 s
--------------------------------------------------------------

Modified catkin_pkg:
Run 1: Found 1460 packages in 0.3 s
Run 1: Topological order took 0.9 s
Run 2: Found 1460 packages in 0.3 s
Run 2: Topological order took 0.9 s
--------------------------------------------------------------
```

src/catkin_pkg/topological_order.py
seeraven: Member renamed.

src/catkin_pkg/topological_order.py
seeraven: Yes that was because the method added its own package first. I've changed it now to store the own package name not in the cache variable, and add the own package name to the given dependencies explicitly.

dirk-thomas: After reviewing the patch again I notices that `depends_for_topological_order` also contains its own package name (which wasn't obvious from the comments / docblock). So I guess it is just consequential that `_recursive_run_depends_for_topological_order` can do the same.

I modified the patch to keep `self.package.name` in the cached member as you had it before but also added / updated the comments / docblock to reflect this detail: 6b2ac587aa437c16ed5969b7eb68c6d2b10274cb.

dirk-thomas:
Thank you for the detailed benchmark - the improvements for larger workspace are great!

I also double checked that the actually returned list of ordered packages is the same before and after the patch - looking good.

Issue #281, opened 2020-04-13
Condition can't handle multiple sequential 'or's


MatthijsBurgh:
The following condition results in a list of length 5
```
condition = "$ROS_PYTHON_VERSION == 2 or $ROS_PYTHON_VERSION == 2 or $ROS_PYTHON_VERSION == 2"
```
It creates the following AssertionError, as it only accepts a list of 3.
```
In [18]: evaluate_condition(condition, os.environ)
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-18-68713d7cacd4> in <module>()
----> 1 evaluate_condition(condition, os.environ)

/usr/lib/python2.7/dist-packages/catkin_pkg/condition.pyc in evaluate_condition(condition, context)
     27         raise ValueError(
     28             "condition '%s' failed to parse: %s" % (condition, e))
---> 29     return _evaluate(parse_results.asList()[0], context)
     30 
     31 

/usr/lib/python2.7/dist-packages/catkin_pkg/condition.pyc in _evaluate(parse_results, context)
     59 
     60     # recursion
---> 61     assert len(parse_results) == 3
     62 
     63     # handle logical operators

AssertionError: 
```
This is the actual result, with a length of 5.
```
[['$ROS_PYTHON_VERSION', '==', '2'],
 'or',
 ['$ROS_PYTHON_VERSION', '==', '2'],
 'or',
 ['$ROS_PYTHON_VERSION', '==', '2']]
```

dirk-thomas:
This should be fixed by #279 which will be part of the next patch release.

PR #283, opened 2020-04-13
use parse actions to evaluate conditions

### LINKED ISSUE #277: Logical operators require parentheses around operands

MatthijsBurgh:
Inspired by https://github.com/pyparsing/pyparsing/blob/master/examples/simpleBool.py
With the use of ParseActions, the result are specific classes in which
better use is made of the results of the operatorPrecedence.

I wanted to use `bool` with an extra argument, but that isn't allowed, therefore I chose to implement `__call__` for all the classes.

Fixes #277, #281, improvement of #279

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
-os: linux
 dist: xenial
+sudo: false
 python:
   - "2.7"
   - "3.4"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
+os: linux
 dist: xenial
-sudo: false
 python:
   - "2.7"
   - "3.4"
--- a/src/catkin_pkg/condition.py
+++ b/src/catkin_pkg/condition.py
@@ -14,64 +14,38 @@
 
 import operator
 
-import pyparsing
+import pyparsing as pp
 
 
-def evaluate_condition(condition, context):
-    if condition is None:
-        return True
-    expr = _get_condition_expression()
-    try:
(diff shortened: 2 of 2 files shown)


MatthijsBurgh:
@dirk-thomas I created this before I found out that a fix was already implemented in master, but not yet released.

I think this is a more elegant solution as we don't abuse `_evaluate` to be a recursive function which handles all cases all parts of the solution. Instead create different classes with each handling there part of the solution.

Please let me know what you think of it.

dirk-thomas:
Thanks for this improvement.

I updated the title to reflect the actual content of the patch. I also added some commits to mainly keep the diff more readable / minimal. I hope they make sense to you.

* revert unrelated change to .travis.yml
* move evaluate_condition() function to the original location to minimize diff
* move new classes to the end in order of usage
* prefix them with an underscore to avoid making them public API

MatthijsBurgh:
@dirk-thomas looks good!

PR #285, opened 2020-04-22
bump min CMake version to avoid CMP0048 warning

mikaelarguedas:
So that newly created packages dont run into https://github.com/ros/catkin/pull/1052


dirk-thomas:
Thanks for the patch.

PR #282, opened 2020-04-13
Fix crash with pyparsing 3

rotu:
pyparsing.operatorPrecedence was deprecated in pyparsing 1.5.7 and removed in pyparsing 3

https://github.com/pyparsing/pyparsing/blob/master/CHANGES
```
colcon build --packages-up-to rmw_cyclonedds_cpp
[8.139s] ERROR:colcon.colcon_core.package_identification:Exception in package identification extension 'ros.bazel' in '/opt/ros/master/src/osrf/osrf_pycommon': module 'pyparsing' has no attribute 'operatorPrecedence'
Traceback (most recent call last):
  File "/Users/dan/Documents/colcon_ws/src/colcon-core/colcon_core/package_identification/__init__.py", line 143, in _identify
    retval = extension.identify(_reused_descriptor_instance)
  File "/Users/dan/Documents/colcon_ws/src/colcon-ros-bazel/colcon_ros_bazel/package_identification/ros_bazel.py", line 39, in identify
    ros_extension.identify(tmp_desc)
  File "/Users/dan/Documents/colcon_ws/src/colcon-ros/colcon_ros/package_identification/ros.py", line 59, in identify
    pkg, build_type = get_package_with_build_type(str(desc.path))
  File "/Users/dan/Documents/colcon_ws/src/colcon-ros/colcon_ros/package_identification/ros.py", line 163, in get_package_with_build_type
    pkg = _get_package(path)
  File "/Users/dan/Documents/colcon_ws/src/colcon-ros/colcon_ros/package_identification/ros.py", line 195, in _get_package
    pkg.evaluate_conditions(os.environ)
  File "/usr/local/Cellar/python/HEAD-29356e0/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/catkin_pkg/package.py", line 212, in evaluate_conditions
    conditional.evaluate_condition(context)
  File "/usr/local/Cellar/python/HEAD-29356e0/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/catkin_pkg/package.py", line 369, in evaluate_condition
    self.evaluated_condition = evaluate_condition(self.condition, context)
  File "/usr/local/Cellar/python/HEAD-29356e0/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/catkin_pkg/condition.py", line 23, in evaluate_condition
    expr = _get_condition_expression()
[...]

PR #287, opened 2020-05-09
Restore compatibility with pyparsing < 1.5.7

cottsay:
RHEL 7 currently has pyparsing 1.5.6.

Compatibility was regressed by #282

### DIFF
--- a/src/catkin_pkg/condition.py
+++ b/src/catkin_pkg/condition.py
@@ -16,6 +16,12 @@ import operator
 
 import pyparsing as pp
 
+# operatorPrecedence renamed to infixNotation in 1.5.7
+try:
+    from pyparsing import infixNotation as infixNotation
+except ImportError:
+    from pyparsing import operatorPrecedence as infixNotation
+
 
 def evaluate_condition(condition, context):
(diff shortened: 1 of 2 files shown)


cottsay:
If this was just one of the tools, I'd be inclined to just drop Python 2 support and make the Python 3 package the default, but `catkin_pkg` is a dependency of ROS itself. Regressing this compatibility made it impossible to build Melodic using the system packages.

src/catkin_pkg/condition.py
```diff
@@ -16,6 +16,12 @@
 
 import pyparsing as pp
 
+# operatorPrecedence renamed to infixNotation in 1.5.7
+try:
+    from pyparsing import infixNotation as infixNotation
```
dirk-thomas, line 21: Nitpick: no need for `as infixNotation`.

dirk-thomas: Removed in 47fe50f46c3d2ae71d46bb149eaf1a92de5e9007.

rotu:
Thank you @cottsay!

PR #291, opened 2020-05-13
catkin_generate_changelog: hack to allow tag version prefixed with v

flixr:
This is not pretty but might help some other people who also want to use `catkin_generate_changelog` for repos where tags have a `v` prefix.

### DIFF
--- a/src/catkin_pkg/changelog_generator.py
+++ b/src/catkin_pkg/changelog_generator.py
@@ -87,16 +87,16 @@ def get_forthcoming_changes(vcs_client, skip_merges=False):
 def _get_version_tags(vcs_client):
     # get all tags in descending order
     tags = vcs_client.get_tags()
-    version_tags = [t for t in tags if re.match(r'^\d+\.\d+.\d+$', t.name)]
+    version_tags = [t for t in tags if re.match(r'^v?\d+\.\d+.\d+$', t.name)]
     return version_tags
 
 
 def _get_latest_version_tag_name(vcs_client):
     # get latest tag
     tag_name = vcs_client.get_latest_tag_name()
-    if not re.match(r'^\d+\.\d+.\d+$', tag_name):
+    if not re.match(r'^v?\d+\.\d+.\d+$', tag_name):
         raise RuntimeError(
-            "The tag name '{}' doesn't match the version pattern x.y.z".format(tag_name))
+            "The tag name '{}' doesn't match the version pattern v?x.y.z".format(tag_name))
     return tag_name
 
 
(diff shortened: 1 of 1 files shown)


src/catkin_pkg/changelog_generator.py
```diff
@@ -218,14 +218,17 @@ def replace_section(match):
     return data if count == 1 else None
 
 
+def version_from_tag(tag_name):
+    return tag_name.replace('v', '') if tag_name is not None else None
```
dirk-thomas, line 222: Please modify this logic to only match the specific pattern accepted elsewhere: the `v` needs to be the first character and shouldn't be anywhere else in the string.

130s:
What about [REP-0132](https://www.ros.org/reps/rep-0132.html) that defines changelog's format?

> The version number consists of three positive integers separate by single dots, e.g. 1.2.3 as specified in [[8]](https://www.ros.org/reps/rep-0132.html#id22).

dirk-thomas:
> What about REP-0132 that defines changelog's format?

@130s Can you please include more context. At least to me it is unclear what you are trying to say.

src/catkin_pkg/changelog_generator.py
dirk-thomas: @flixr Friendly ping.

flixr: done

130s:
> > What about REP-0132 that defines changelog's format?
> 
> @130s Can you please include more context. At least to me it is unclear what you are trying to say.

I may be wrong, but my understanding is:
- REP-0132 specifies the package version format to be "version number consists of three positive integers separate by single dots, e.g. 1.2.3 as specified in [[8]](https://www.ros.org/reps/rep-0132.html#id22)."
- `catkin_generate_changelog` conforms to REP-0132.

Over the years I've seen developers trying to use version name format that doesn't follow the mentioned syntax. So if there's a room for accommodating custom version format in catkin, I'm +1. If so, I'd just have to suggest to update REP-0132 to clarify that the custom format can be used.

130s:
Separate argument is that allowing "v" prefix looks very specific. Unless there is a widely accepted versioning scheme that uses "v", it may make more sense to change the functionality so that a user can pass their own custom prefix IMO.

dirk-thomas:
This patch is just focused on being able to collect the changelog for tags with a `v` prefix. It shouldn't affect the standard workflow so I don't mind supporting it.

REP 132 only defines the format of a version number in each changelog section. It neither specifies how the information is retrieved from a repository nor mentions `catkin_generate_changelog`. Therefore I don't see how it should / needs to be updated.

There no promise whatsoever that using tags in the form of `vX.Y.Z` is supported in any other part of the toolchain. We are happy to take contributions if anyone would like to provide additional flexibility (assuming it is well thought through and has no conceptional or other downsides).

130s:
@dirk-thomas Then, do I understand correctly that catkin (not catkin_generate_changelog) only supports x.y.z version? If so, I can see developers would be confused when catkin_generate_changelog allows custom version format but catkin doesn't.

dirk-thomas:
> do I understand correctly that catkin (not catkin_generate_changelog) only supports x.y.z version?

What exact part in `catkin` are you referring to? The package manifest is specified in another REP and implemented in `catkin_pkg`. I am not sure what else in `catkin` considers version numbers.

flixr:
We tag all all releases with a v prefix and this is also already supported
by bloom. As @dirk-thomas said that does not mean that the rep should also
support this in the changelog.

Dirk Thomas <notifications@github.com> schrieb am Di., 23. Juni 2020, 19:33:

PR #292, opened 2020-05-15
add GitLab auto linking for issues in catkin_generate_changelog

patrickcjh:


dirk-thomas:
Thank you for the contribution.

patrickcjh:
Thanks for accepting it.

PR #294, opened 2020-08-10
Compare dependencies with all attributes

f-fl0:
## Problem statement

Running `catkin_make` in a workspace with a package with dependencies like this:
```
<depend condition="$ROS_DISTRO == kinetic" version_eq="1.9.3-r0">diagnostic_updater</depend>
<depend condition="$ROS_DISTRO == noetic" version_eq="1.9.4-r0">diagnostic_updater</depend>
```
will result in the following error:
```
Error(s):
- The generic dependency on 'diagnostic_updater' is redundant with: build_depend, build_export_depend, exec_depend
```

## Proposed solution

Comparing dependencies using their names and attributes (e.g. conditions) instead of just their names.

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -672,9 +672,9 @@ def parse_package_string(data, filename=None, warnings=None):
         depends = _get_dependencies(root, 'depend')
         for dep in depends:
             # check for collisions with specific dependencies
-            same_build_depends = ['build_depend' for d in pkg.build_depends if d.name == dep.name]
-            same_build_export_depends = ['build_export_depend' for d in pkg.build_export_depends if d.name == dep.name]
-            same_exec_depends = ['exec_depend' for d in pkg.exec_depends if d.name == dep.name]
+            same_build_depends = ['build_depend' for d in pkg.build_depends if d == dep]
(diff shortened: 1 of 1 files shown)


dirk-thomas:
While considering more attributes than `name` is certainly the right thing to do I am concerned about including `evaluated_condition` (see https://github.com/ros-infrastructure/catkin_pkg/blob/e7a6806b370c9318ce8a792149c8906830ba22e6/src/catkin_pkg/package.py#L324) in this comparison. Both instances could have the same attributes and only that computed attribute could be different, I would think that those should still be considered duplicates. Maybe the right thing to do would be to exclude `evaluated_condition` from `__eq__`?

f-fl0:
> While considering more attributes than `name` is certainly the right thing to do I am concerned about including `evaluated_condition` (see
> 
> https://github.com/ros-infrastructure/catkin_pkg/blob/e7a6806b370c9318ce8a792149c8906830ba22e6/src/catkin_pkg/package.py#L324
> ) in this comparison. Both instances could have the same attributes and only that computed attribute could be different, I would think that those should still be considered duplicates. Maybe the right thing to do would be to exclude `evaluated_condition` from `__eq__`?

@dirk-thomas `__eq__` now ignores `evaluated_condition` as you suggested. Let me know if you have other comments.

dirk-thomas:
Thanks for the patch.

PR #295, opened 2020-08-29
support quoted values in conditions

dirk-thomas:
The patch enables to express a condition like this: `$FOO == ""`.

This requires REP 149 to be updated (or if that is not feasible a new REP / format version). **Update:** see ros-infrastructure/rep#288.


ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/rfc-on-update-of-rep-149-package-format-3-conditions-with-quotes-string-literals/16370/1

gbiggs (approved):

PR #296, opened 2020-09-18
Validate license values

shr-project:


src/catkin_pkg/package.py
```diff
@@ -222,6 +222,153 @@ def validate(self, warnings=None):
...
         new_warnings = []
 
+        def is_valid_spdx_identifier(lic):
+            """
+            Check if the license is already one of valid SPDX Identifiers.
+
+            The list was created from https://spdx.org/licenses/ with:
+            cat doc/spdx-3.10-2020-08-03.csv | cut -f 2 | grep -v ^Identifier$
+            """
+            return lic in ['0BSD', 'AAL', 'Abstyles', 'Adobe-2006', 'Adobe-Glyph', 'ADSL', 'AFL-1.1', 'AFL-1.2', 'AFL-2.0', 'AFL-2.1', 'AFL-3.0', 'Afmparse', 'AGPL-1.0-only', 'AGPL-1.0-or-later',
```
130s, line 232: I don't have an answer, but I'm afraid embedding a (huge) list of license names can be a maintenance burden (I saw you described how you made it, but still would require manual invocation in the future when updating the list is needed. Also as a disclaimer, I'm not a maintainer on this repo). Wish there's a standalone tool that provides such a list.

130s (commented):
Great tool! I was/am irritated by license values inconsistently named when I'm working on things like https://github.com/ros-infrastructure/rospkg/pull/201. I like the direction of the feature is going. I'm not a maintainer and not so sure about the direction of the implementation though.

cottsay:
This seems like it might be better suited for inclusion in [catkin_lint](https://github.com/fkie/catkin_lint). Any thoughts on that?

src/catkin_pkg/package.py
shr-project: Well, my hope is that this list is only temporary and component owners will start using valid identifiers in package.xml and eventually only the validation function will stay and all the replacements won't be needed. So the only maintenance should be to drop it in some far future.

shr-project:
> This seems like it might be better suited for inclusion in [catkin_lint](https://github.com/fkie/catkin_lint). Any thoughts on that?

I've added it to catkin_pkg only because rosdistro CI github action uses it to validate version, so I've used the same mechanism for license validation, but if more people will notice the warnings from catkin_lint then definitely it should be there (as well).

src/catkin_pkg/package.py
shr-project: Also the list isn't complete (and cannot be), because on of the most common issues is using license like LGPL without a version, but here without checking the actual component source we cannot replace it with anything more accurate, all we can do is to show and warning and hope that the owner will improve the package.xml.

I still plan to extend the list to cover most common cases of listing multiple licenses in one <license> tag instead of using multiple of them - just using & as separator - together with slightly different warning message suggesting to split them into multiple tags.

gavanderhoorn: > Well, my hope is that this list is only temporary and component owners will start using valid identifiers in package.xml and eventually only the validation function will stay and all the replacements won't be needed. So the only maintenance should be to drop it in some far future.

that would be great, but I would say this is not something we can expect (ie, the: "component owners will start using valid identifiers in package.xml and eventually only the validation function will stay and all the replacements won't be needed" part).

As of today, there are 5946 packages in the `ros/rosdistro` db alone, with many more not registered anywhere (ie: hosted somewhere publicly but not part of a distribution). Many of those packages are old and will not be updated, but even the more recent ones will have a significant nr which will not be updated.

If you add this list, I expect it to be necessary for at least until Noetic EOLs, and probably also for a few ROS 2 versions (if you intend to add this functionality to a ROS 2 tool as well). We're talking 5+ years here.

shr-project: Yes, that's the far future for removal, but in the meantime we shouldn't need to update the list.

The replacements are just to have some immediate improvements where possible, while showing the corresponding SPDX in warning message. And not to regress from what superflore currently does with regexp replacements.

It's not meant to be perfect and it cannot be without parsing the actual source files with something like scancode-toolkit, because in some cases the valid identifier in package.xml doesn't match with the actual license and catkin_pkg nor superflore could detect that (even to show another warning).

dirk-thomas:
@shr-project Atm the PR is in draft state and has no description. Therefore I am not going to review the patch. Please add a very detailed description about what the patch is doing, why it is doing so, what the side effects are, why you think those are justified, etc.

On a first look this adds new warnings to **a lot** of existing packages. This on its own for already released distros is an almost guaranteed no-go.

shr-project:
> @shr-project Atm the PR is in draft state and has no description. Therefore I am not going to review the patch. Please add a very detailed description about what the patch is doing, why it is doing so, what the side effects are, why you think those are justified, etc.

It was in draft state, because I wanted to add separate warning for license values which include multiple licenses (instead of using multiple separate license tags). I've added another commit for that now and added description.

Please let me know if it makes more sense to you now.

> On a first look this adds new warnings to a lot of existing packages. This on its own for already released distros is an almost guaranteed no-go.

Are developers used to use https://github.com/fkie/catkin_lint ? Would showing warning about strange license texts in components from released distros be OK if done in catkin_lint?

cottsay:
> Would showing warning about strange license texts in components from released distros be OK if done in catkin_lint?

It's certainly not a hill I'm willing to die on, but in my opinion, the warning and error messages coming out of `catkin_pkg` should be more along the lines of syntax, structure, or schema-level problems in the manifests - problems that could reasonably prevent accurate processing of the manifest by `catkin_pkg` itself, or another system that depends on `catkin_pkg`. It isn't a perfect parallel, but I'll note that many Linux distributions document what precise representation should be used to refer to specific licenses, but few enforce this in the packaging system itself (i.e. deb/rpm as opposed to linters and test automation).

To be clear: I think this is an excellent thing to do, I'm just not convinced `catkin_pkg` is the right tool for the job. I would _love_ to see it in `catkin_lint`, and I would _love_ to see widespread adoption of `catkin_lint` in a GitHub Action.

shr-project:
I'm working on very similar PR for catkin_lint as well now :).

I only work with ROS when working on meta-ros, I've never used ROS as a regular developer, so I don't know what's the usual work flow for people who write package.xml. If it's common that they run catkin_lint (or that Github Actions runs it for them), then I agreed that's definitely the place where it should be validated.

gavanderhoorn:
@cottsay wrote:
>  I would _love_ to see it in `catkin_lint`, and I would _love_ to see widespread adoption of `catkin_lint` in a GitHub Action.

`industrial_ci` supports `catkin_lint` runs natively and can be used as a GH Action.

@shr-project wrote:
> If it's common that they run catkin_lint

can't speak for everyone, but I have quite a few `industrial_ci` configurations which have `CATKIN_LINT=true`.

PR #301, opened 2020-12-09
Add GitHub Actions workflow for catkin_pkg CI.

nuclearsandwich:
Travis CI is becoming increasingly unavailable. This ports the same test suite to run via GitHub Actions based on the workflow I used for rosdep. It does add additional operating systems and python versions to the test matrix.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,53 @@
+name: catkin_pkg-ci
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: ['*']
+
+jobs:
+    build:
+      strategy:
+        matrix:
+          os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04, macos-latest]
+          python: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
+          exclude:
+          - os: ubuntu-16.04
+            python: 3.6
+          - os: ubuntu-16.04
+            python: 3.7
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: ['*']
+
+jobs:
+    build:
+      strategy:
+        matrix:
```
nuclearsandwich, line 12: @cottsay With apologies that I'm experimenting rather than being consistent still I'd love your feedback on this approach to the build matrix. It's the smallest I could make it while still having any coherence. If you think this is more readable and maintainable than the big list of exclusions I'll open PRs to normalize it everywhere.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,41 @@
...
+          - os: ubuntu-16.04
+            python: 2.7
+          - os: ubuntu-18.04
+            python: 2.7
+          - os: macos-latest
+            python: 2.7
+          - os: ubuntu-16.04
+            python: 3.5
+          - os: ubuntu-18.04
+            python: 3.6
```
nuclearsandwich, line 25: @cottsay With apologies that I'm experimenting rather than being consistent still I'd love your feedback on this approach to the build matrix. It's the smallest I could make it while still having any coherence. If you think this is more readable and maintainable than the big list of exclusions I'll open PRs to normalize it everywhere.

PR #302, opened 2021-01-15
Address flake8 blind Exception errors.

SidFaber:
Update six try/except blocks to catch specific and expected errors. The release of flake8-blind-except v0.2.0 on1/7/2021 (https://github.com/elijahandrews/flake8-blind-except/releases/tag/v0.2.0) caused flake8 build errors on these blocks.

Also add three tests for improperly formatted package XML.

Signed-off-by: Sid Faber <sid.faber@canonical.com>

### DIFF
--- a/test/test_metapackage.py
+++ b/test/test_metapackage.py
@@ -62,11 +62,7 @@ def assert_warning(warnreg):
 
 
 def _validate_metapackage(path, package):
-    try:
-        validate_metapackage(path, package)
-    except Exception:
-        # print('on package ' + package.name, file=sys.stderr)
-        raise
+    validate_metapackage(path, package)
 
 
 class TestMetapackageValidation(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


test/test_metapackage.py
```diff
@@ -62,11 +62,7 @@ def assert_warning(warnreg):
 
 
 def _validate_metapackage(path, package):
-    try:
-        validate_metapackage(path, package)
-    except Exception:
-        # print('on package ' + package.name, file=sys.stderr)
-        raise
+    validate_metapackage(path, package)
```
nuclearsandwich, line 65: With the meat of this wrapper function removed does it make sense to remove it entirely and call validate_metapackage directly in the tests?

SidFaber: Yeah, that makes a lot of sense. Updated with 9eab131c6193024ddb48750d384c3b96b47f60e6, thanks!

PR #305, opened 2021-04-01
Explicitly require setuptools, changelog.py imports pkg_resources

hrnciar:


cottsay (approved):
Thanks!

For review purposes, here's the spot: https://github.com/ros-infrastructure/catkin_pkg/blob/e5f553826cff83b18cf2d438c7df6c13b3c95c4a/src/catkin_pkg/changelog.py#L53

PR #306, opened 2021-08-10
Generate documentation using sphinx-rtd-theme.

nuclearsandwich:
These two commands, together with forthcoming changes for the doc-independent job configuration will switch the catkin_pkg API documentation to the rtd theme.

So we're not as consistent currently as I thought in the offline discussion earlier today.
rospkg and rosdep currently use the ros-theme while rosinstall and vcstools currently use the haiku theme. catkin-pkg and rosdistro use the default sphinx theme since both generate their sphinx conf.py files with sphinx-apidoc. 

Trying to apply retroactive logic to the above, it seems like ros utility packages use the ROS theme and vcstools packages (including rosinstall) use the haiku theme, while catkin_pkg and rosdistro, which have a conf.py generated by sphinx-apidoc, use the default theme.

The scripts here could easily be modified to target the ROS theme rather than RTD although if I compare the two I personally find rtd to be more readable. Once this is settled I'll open a sibling PR on rosdistro so that the two of them are using the same theme.


doc/Makefile
```diff
@@ -7,6 +7,8 @@ html: apidoc
 
 apidoc:
 	$(SPHINXAPIDOC) -F -o _build ../src/catkin_pkg
+	sed -i "/extensions =\[/a 'sphinx_rtd_theme'," _build/conf.py
+	sed -i "s|html_theme =.*$$|html_theme = 'sphinx_rtd_theme'|" _build/conf.py
```
wjwwood, line 11: Hmm, this works but seems pretty fragile. If the `conf.py` that is generated changes it might break this, no?

Two other options are:

- the template dir option https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html
- make a `conf.py` and add `sphinx-autodoc` as an extension, e.g. https://eikonomega.medium.com/getting-started-with-sphinx-autodoc-part-1-2cebbbca5365

The benefit of using it how it is now is that you don't need the (mostly) boilerplate conf file and index file. But now we're trying to get that benefit while also changing something about the conf, which I think is tenuous.

That all being said, I'm sure this works right now.

What do you think?

wjwwood: Sorry, I didn't get back sooner.

To be clear, I'm fine with the current approach, we just might have to do something with it later, but that's ok.

I think the idea of the templates directory option is that you provide templates for things like the `conf.py` which `sphinx-apidoc` will use instead of the standard boilerplate. I think it's completely separate from the idea of a theme or templates used to say generate html headers and footers. It might be equally fragile thought, to be honest.

> Is the creation of a conf.py and index the only difference between sphinx-apidoc and sphinx with the autodoc extension?

Indeed, if you look at what `sphinx-apidoc` does, it just generates a `conf.py` and `index.rst` and runs `sphinx-build` over it (more or less).

The link I provided for this option basically walks you through creating your own `conf.py` and `index.rst` to accomplish this yourself.

PR #307, opened 2021-08-19
Make filenames to be used as ignore markers configurable

janstrohbeck:
catkin_pkg ignores packages that include one of {CATKIN_IGNORE, COLCON_IGNORE, AMENT_IGNORE}. For catkin_tools (which utilizes functions from catkin_pkg), I would like to implement a behavior such that it will only ignore packages that include CATKIN_IGNORE (and not those that have COLCON_IGNORE/AMENT_IGNORE). This PR does not change any functionality, it only makes the set of ignore markers configurable via a keyword argument. Link to catkin_tools PR: https://github.com/catkin/catkin_tools/pull/688

### DIFF
--- a/src/catkin_pkg/packages.py
+++ b/src/catkin_pkg/packages.py
@@ -40,11 +40,14 @@ from .package import PACKAGE_MANIFEST_FILENAME
 from .package import parse_package_string
 
 
-def find_package_paths(basepath, exclude_paths=None, exclude_subspaces=False):
+DEFAULT_IGNORE_MARKERS = {'AMENT_IGNORE', 'CATKIN_IGNORE', 'COLCON_IGNORE'}
+
+
+def find_package_paths(basepath, exclude_paths=None, exclude_subspaces=False, ignore_markers=DEFAULT_IGNORE_MARKERS):
     """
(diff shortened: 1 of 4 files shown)


nuclearsandwich:
At first I wasn't too bothered by this PR and I even just pushed up some basic test cases exercising it. But as I was getting ready to approve I started wondering what use cases there are for making this configurable. The inciting case was helpfully described in the [accompanying catkin_tools pr](https://github.com/catkin/catkin_tools/pull/688) which involves using a shared workspace with ROS 1 and ROS 2 packages and I'm not sure that I could ever endorse or recommend sharing a source space with two different build tools like catkin_tools and colcon. Once tests and linters are passing I think I default to accepting this change because it doesn't break anything in the default tooling but I'd like to get other opinions as well.

nuclearsandwich (approved):
Approved with some hesitation about whether switching any existing tooling to use this option or the enabled use case of running a mixed ROS and ROS 2 source workspace with different build tools should even be encouraged.

cottsay (approved):
I'm lukewarm about whether your driving use case is actually a good idea, but this PR does no harm and provides API flexibility that could be useful.

PR #317, opened 2021-12-02
Update release suites.

nuclearsandwich:
* Drop Ubuntu Xenial and surrounding non-LTS distributions.
* Drop Debian Jessie and Debian Stretch. (Official security support for
  Debian Stretch ended in July 2020 and Jessie even earlier.)
* Add Debian Bullseye for Python 3.
* Add Ubuntu Jammy for Python 3.

Although the non-LTS Ubuntu releases between Bionic and Focal are no
longer supported I have not removed them as the pattern I've observed
the most is not to remove these distributions until removing the
surrounding LTS distribution. However, if pressed I would not fight to
keep those suites in the release list.


nuclearsandwich:
The CI error is macOS related and resolved by https://github.com/ros-infrastructure/catkin_pkg/pull/318

cottsay (approved):

PR #321, opened 2022-01-02
Use unittest.mock where possible

cottsay:
The 'unittest.mock' module has been present since Python 3.3, and some platforms are now dropping the 'mock' backport package.


nuclearsandwich (approved):
We could also consider dropping `mock` installation in the CI workflow as mentioned in https://github.com/ros-infrastructure/rosdep/pull/850#issuecomment-1022434927

PR #323, opened 2022-01-25
Declare test dependencies in [test] extra

cottsay:
Formally declaring the test dependencies somewhere is a good idea, and allows tools like `colcon` to discover them.

['extras_require']['test'] is where we put these dependencies in colcon itself, but there doesn't appear to be a strong consensus among the packaging community regarding where to list these, and many options aren't compatible with python 2, which is still supported in this package. This option seems to meet all of our needs.

Also make the 'mock' dependency conditional, now that this package supports using unittest.mock.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,8 +27,7 @@ jobs:
           python-version: ${{matrix.python}}
       - name: Install dependencies
         run: |
-          python -m pip install argparse coverage docutils mock nose pyparsing python-dateutil
-          python -m pip install flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated flake8-docstrings flake8-import-order flake8-quotes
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
> ['extras_require']['test'] is where we put these dependencies in colcon itself, but there doesn't appear to be a string consensus among the packaging community regarding where to list these, and many options aren't compatible with python 2, which is still supported in this package.

Presumably the proposed `['extras_require']['test']` is compatible as our test on Ubuntu 18.04 with python 2.7 succeed with this PR?

cottsay:
> Presumably the proposed `['extras_require']['test']` is compatible as our test on Ubuntu 18.04 with python 2.7 succeed with this PR?

Indeed it is. I also tested it specifically with the system's python interpreter and setuptools package in Bionic, and it seems to work just fine.

nuclearsandwich (approved):

PR #325, opened 2022-02-16
Enable Python 3.10 tests, bump actions/setup-python

cottsay:


nuclearsandwich (approved):
I find the lack of an upstream changelog for actions/setup-python pretty frustrating since I'd like to know what changes are in v2 compared to v1 but since the configuration works as-is I don't see a reason to block on that.

PR #326, opened 2022-03-02
Use only first line of description in distutils setup generation

cottsay:
This is an alternative to #316 that uses only the first line of the package description in the Python distutils setup rather than converting line breaks to spaces.

I'd like to hear feedback on which approach is preferable.

### DIFF
--- a/src/catkin_pkg/python_setup.py
+++ b/src/catkin_pkg/python_setup.py
@@ -99,13 +99,13 @@ def generate_distutils_setup(package_xml_path=os.path.curdir, **kwargs):
     elif package.urls:
         data['url'] = package.urls[0].url
 
-    description = package.description.splitlines()[0]
+    description = package.plaintext_description.splitlines()[0]
     if len(description) > 200:
         description = description[:197] + '...'
 
     data['description'] = description
-    if description != package.description:
-        data['long_description'] = package.description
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):
I prefer this approach compared with both #316 and the base branch implementation. It has a much better chance of producing an intelligible description within the restrictions of the setuptools description format.

I wonder if it's worth adding any kind of style convention to the package creation documentation espousing the `Subject line\n\nLong description\n` pattern but it shouldn't block this PR.

cottsay:
Depends on #329

cottsay:
Okay, this is looking pretty good. For example:
Before:
```
Summary: The actionlib stack provides a standardized interface for     interfacing with preemptable tasks. Examples of this include moving     the base to a target location, performing a laser scan and retu...
```
After:
```
Summary: The actionlib stack provides a standardized interface for interfacing with preemptable tasks. Examples of this include moving the base to a target location, performing a laser scan and returning th...
```

Dropping a `<br/>` after the first sentence makes this even better:
```
Summary: The actionlib stack provides a standardized interface for interfacing with preemptable tasks.
```

nuclearsandwich (approved):
Updating my approval to the version of this based on the plaintext_description attribute.

PR #329, opened 2022-03-11
Add plaintext description field to Package.

nuclearsandwich:
This patch adds a `_get_node_text` function to the package module's xml processing toolset: a basic plaintext renderer for xhtml descriptions which also extracts text from formatting tags such as `<em>` or `<a>` although it does not render or preserve the href attribute on a tags in any way. All literal whitespace is compressed into a single ` ` so the only way to insert a newline into the plaintext description is by using a `<br>` tag. 

I used this to add a new `plaintext_description` field to a parsed package where I take the extra step of trimming whitespace around newlines even though they are technically valid it doesn't match the expectations established by first writing an xml description and then hand-rendering it as plaintext.

## Motivation and background

There are several cases where a package.xml description is meant to be "rendered" as plain text.
According to [REP-149](https://www.ros.org/reps/rep-0149.html#description) the description "may contain XHTML" but depending on where the description is used "XML tags and multiple whitespaces might be stripped".

The existing description field does not do any sort of whitespace manipulation, nor does it actually strip XHTML tags. This results in extremely odd formatting when used in plaintext contexts, such as a `PKG-INFO` file within a python sdist.

Consider the [description for actionlib](https://github.com/ros/actionlib/blob/3073509690976a8816636d41f04c013633f1c119/actionlib/package.xml#L6-L11)

```
  <description>
    The actionlib stack provides a standardized interface for
    interfacing with preemptable tasks. Examples of this include moving
    the base to a target location, performing a laser scan and returning
    the resulting point cloud, detecting the handle of a door, etc.
  </description>
```

This is currently parsed as
```
'The actionlib stack provides a standardized interface for\n    interfacing with preemptable tasks. Examples of this include moving\n    the base to a target location, performing a laser scan and returning\n    the resulting point cloud, detecting the handle of a door, etc.'
```

All interior whitespace is retained which results in an oddly formatted description since the line breaks do not have any inherent semantic value but appear to be formatted just to adhere to 72-character line widths.

When discussing #316 and #326 both of which are proposals to manipulate the description in order to conform to changes in setuptools each proposal assumed that lines would be "sensibly formatted" to start with and that taking either the first "line" or replacing line breaks with single spaces would retain a sensibly formatted, if truncated description. However if we took only the first line of the above description (as in #326) it would yield:
```
The actionlib stack provides a standardized interface for
```

If we we compressed newlines into spaces and then took the first 200 characters (as in #316) it would yield:
```
The actionlib stack provides a standardized interface for     interfacing with preemptable tasks. Examples of this include moving     the base to a target location, performing a laser scan and retu...
```

While trying to decide which of these is the "lesser of two unfortunates" I got it into my head that we should really be treating these like an HTML renderer does and treating any combination of whitespace as a single space. I turned to the XML specification to back me up on this. From section [2.10 on white space handling](https://www.w3.org/TR/2008/REC-xml-20081126/#sec-white-space)

> An XML processor MUST always pass all characters in a document that are not markup through to the application. A validating XML processor MUST also inform the application which of these characters constitute white space appearing in element content.

Which prompts me to conclude that catkin_pkg is our application and that we should be deciding how to treat whitespace. The description in REP-149 specifically states that whitespace may be stripped in certain situations and I think that a plaintext "rendering" of an xhtml description field like actionlib's should result in the following string:

```
The actionlib stack provides a standardized interface for interfacing with preemptable tasks. Examples of this include moving the base to a target location, performing a laser scan and returning the resulting point cloud, detecting the handle of a door, etc.
```

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -628,7 +628,7 @@ def parse_package_string(data, filename=None, warnings=None):
 
     # description
     pkg.description = _get_node_value(_get_node(root, 'description', filename), allow_xml=True, apply_str=False)
-    pkg.plaintext_description = _get_node_text(_get_node(root, 'description', filename))
+    pkg.plaintext_description = re.sub(' +(\n+) +', r'\1', _get_node_text(_get_node(root, 'description', filename)), flags=re.MULTILINE)
 
     # at least one maintainer, all must have email
     maintainers = _get_nodes(root, 'maintainer')
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -137,6 +137,9 @@ class Package(object):
             data[attr] = getattr(self, attr)
         return str(data)
 
+    def plaintext_description(self):
+        return self.description
+
     def has_buildtool_depend_on_catkin(self):
         """
         Return True if this Package buildtool depends on catkin, otherwise False.
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -813,7 +813,7 @@ def _get_node_text(node, strip=True):
     value = ''
     for child in node.childNodes:
         if child.nodeType == child.TEXT_NODE:
-            value += re.sub('\s+', ' ', child.data)
+            value += re.sub(r'\s+', ' ', child.data)
         elif child.nodeType == child.ELEMENT_NODE:
             if child.tagName == 'br':
                 value += '\n'
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -137,7 +137,7 @@ class Package(object):
             data[attr] = getattr(self, attr)
         return str(data)
 
-    def plaintext_description(self):
+    def get_plaintext_description(self):
         return self.description
 
     def has_buildtool_depend_on_catkin(self):
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -60,6 +60,7 @@ class Package(object):
         'version',
         'version_compatibility',
         'description',
+        'plaintext_description',
         'maintainers',
         'licenses',
         'urls',
@@ -138,7 +139,7 @@ class Package(object):
         return str(data)
 
     def get_plaintext_description(self):
-        return self.description
+        return self.plaintext_description
 
     def has_buildtool_depend_on_catkin(self):
         """
@@ -627,6 +628,7 @@ def parse_package_string(data, filename=None, warnings=None):
 
     # description
     pkg.description = _get_node_value(_get_node(root, 'description', filename), allow_xml=True, apply_str=False)
+    pkg.plaintext_description = _get_node_text(_get_node(root, 'description', filename))
 
     # at least one maintainer, all must have email
     maintainers = _get_nodes(root, 'maintainer')
@@ -804,6 +806,23 @@ def _get_node_value(node, allow_xml=False, apply_str=True):
         value = (''.join([n.data for n in node.childNodes if n.nodeType == n.TEXT_NODE])).strip(' \n\r\t')
     if apply_str:
         value = str(value)
+    return value
+
+
+def _get_node_text(node, strip=True):
+    value = ''
+    for child in node.childNodes:
+        if child.nodeType == child.TEXT_NODE:
+            value += re.sub('\s+', ' ', child.data)
+        elif child.nodeType == child.ELEMENT_NODE:
+            if child.tagName == 'br':
+                value += '\n'
+            else:
+                value += _get_node_text(child, strip=False)
+        else:
+            assert 'unreachable'
+    if strip:
+        value = value.strip()
     return value
 
 
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -138,9 +138,6 @@ class Package(object):
             data[attr] = getattr(self, attr)
         return str(data)
 
-    def get_plaintext_description(self):
-        return self.plaintext_description
-
     def has_buildtool_depend_on_catkin(self):
         """
         Return True if this Package buildtool depends on catkin, otherwise False.
--- a/test/test_package.py
+++ b/test/test_package.py
@@ -388,7 +388,7 @@ class PackageTest(unittest.TestCase):
         assert package.get_plaintext_description() == """\
 A package with an XHTML description.
 
-This package contains several xhtml tags which are, according to REP-149, meant to be handled properly but "XML tags and multiple whitespaces" may be stripped in some situations. Ahother sentence in this quasi-paragraph will continue to appear on the same plaintext line becuase there was no <br/> tag to indicate a newline should appear in the output text.
+This package contains several xhtml tags which are, according to REP-149, meant to be properly handled but "XML tags and multiple whitespaces" may be stripped in some situations. Another sentence in this quasi-paragraph will continue to appear on the same plaintext line because there was no <br/> tag to indicate a newline should appear in the output text.
 This text should appear on a subsequent line.\
 """
 
--- a/test/test_package.py
+++ b/test/test_package.py
@@ -385,7 +385,7 @@ class PackageTest(unittest.TestCase):
         filename = os.path.join(test_data_dir, 'xhtml_description.xml')
         package = parse_package(filename)
         assert package.description
-        assert package.get_plaintext_description() == """\
+        assert package.plaintext_description == """\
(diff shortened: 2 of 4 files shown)


cottsay (approved):

nuclearsandwich:
Thanks for the initial review @cottsay I ought to've clarified that I planned to do some iterating _in this PR_ rather than just merge an as-yet-unused test fixture. I should have opened this PR in draft and have now flipped it to one.

test/test_package.py
```diff
@@ -381,6 +381,16 @@ def test_parse_package_invalid(self):
...
 
+    def test_parse_package_xhtml_description(self):
+        filename = os.path.join(test_data_dir, 'xhtml_description.xml')
+        package = parse_package(filename)
+        assert package.description
+        assert package.plaintext_description() == """
+A package with an XHTML description.
+
+This package contains several xhtml tags which are, according to REP-149, meant to be handled properly but "XML tags and multiple whitespaces" may be stripped in some situations.
+        """
```
nuclearsandwich, line 392: I'm looking for pre-review feedback. Do folks generally agree that this is an appropriate plaintext rendering of the html description in [`test/data/package/xhtml_description.xml`](https://github.com/ros-infrastructure/catkin_pkg/blob/51ee3aad4b7169f04cbf32d78fb7c02066f23d6f/test/data/package/xhtml_description.xml)?

src/catkin_pkg/package.py
```diff
@@ -137,6 +137,9 @@ def __str__(self):
             data[attr] = getattr(self, attr)
         return str(data)
 
+    def plaintext_description(self):
+        return self.description
+
```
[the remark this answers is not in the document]
cottsay: I'm OK with a function, but I think it should start with `get_` if it is.

test/test_package.py
nuclearsandwich: I believe that everything inside the """ is getting captured, so we shouldn't expect to see the leading or trailing whitespace and line breaks.

I noticed this as well and pushed 4bba520.  Thanks for the close read!

src/catkin_pkg/package.py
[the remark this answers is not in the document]
nuclearsandwich: Done in [4deff0a](https://github.com/ros-infrastructure/catkin_pkg/pull/329/commits/4deff0aeb7366b0b722b5849524607192c7b91b2)

test/test_package.py
nuclearsandwich: > What you have looks right to me. Please include an additional sentence after the last one to explicitly show that behavior as well.

[`a79c056` (#329)](https://github.com/ros-infrastructure/catkin_pkg/pull/329/commits/a79c0564ce67285821e5d8145423d8638bb03122)

nuclearsandwich:
The latest push gets us closer but still not home. There are some small whitespace discrepancies between what I've got and our intended rendering of the text which I need to figure out if I can handle in a reasonable way or if we have to make concessions in our expectations. Incidentally, I moved the processing of the plaintext_description into the xml processing phase since it's much easier to do this on an extant DOM rather than trying to do something at runtime. Since I'm storing this in a `plaintext_description` attribute I think we can probably cut the getter function and rewrite the test to just check the attribute.

```
    """
E   assert ('A package with an XHTML description. \n'\n '\n'\n ' This package contains several xhtml tags which are, according to REP-149, '\n 'meant to be properly handled but "XML tags and multiple whitespaces" may be '\n 'stripped in some situations. Another sentence in this quasi-paragraph will '\n 'continue to appear on the same plaintext line because there was no <br/> tag '\n 'to indicate a newline should appear in the output text. \n'\n ' This text should appear on a subsequent line.') == ('A package with an XHTML description.\n'\n '\n'\n 'This package contains several xhtml tags which are, according to REP-149, '\n 'meant to be properly handled but "XML tags and multiple whitespaces" may be '\n 'stripped in some situations. Another sentence in this quasi-paragraph will '\n 'continue to appear on the same plaintext line because there was no <br/> tag '\n 'to indicate a newline should appear in the output text.\n'\n 'This text should appear on a subsequent line.')
E     - A package with an XHTML description.
E     + A package with an XHTML description.
E     ?                                     +
E
E     - This package contains several xhtml tags which are, according to REP-149, meant to be properly handled but "XML tags and multiple whitespaces" may be stripped in some situations. Another sentence in this quasi-paragraph will continue to appear on the same plaintext line because there was no <br/> tag to indicate a newline should appear in the output text.
E     +  This package contains several xhtml tags which are, according to REP-149, meant to be properly handled but "XML tags and multiple whitespaces" may be stripped in some situations. Another sentence in this quasi-paragraph will continue to appear on the same plaintext line because there was no <br/> tag to indicate a newline should appear in the output text.
E     ? +                                                                                                                                                                                                                                                                                                                                                                     +
E     - This text should appear on a subsequent line.
E     +  This text should appear on a subsequent line.
E     ? +

test/test_package.py:388: AssertionError
```

tfoote (approved):
This looks good to me. There's a level that I don't think that we should do a character by character test for all the parsed outputs. But maybe just a few checks for specific things that are properly escaped etc would make sense. And we can trust that the specific behaviors are well defined by the parser that we're relying upon and we can just spec that we'll follow the parser.

nuclearsandwich:
As much as it _bugs_ me, the result of the current processing _is_ actually, more or less, what an html plaintext rendering would produce. I think I do want to try and strip spaces around newlines but I may not keep it unless I can find a way to prove it doesn't misbehave.

src/catkin_pkg/package.py
```diff
@@ -804,6 +809,23 @@ def _get_node_value(node, allow_xml=False, apply_str=True):
...
+        elif child.nodeType == child.ELEMENT_NODE:
+            if child.tagName == 'br':
+                value += '\n'
+            else:
+                value += _get_node_text(child, strip=False)
+        else:
+            assert 'unreachable'
+    if strip:
+        value = value.strip()
+    return value
```
cottsay, line 826: Maybe something like this could help? As I understand it, the only way for a newline to show up is when `<br/>` is used, so this should hold for all cases.
```suggestion
    return re.sub('\s?\n\s?', '\n', value)
```

nuclearsandwich: This is actually really close to what I ended up with in https://github.com/ros-infrastructure/catkin_pkg/pull/329/commits/b251bb7ee1914d16a443b0d93639dfe51b307872. I implemented it in _only_ for the plaintext_description as I don't think it's generally "less correct", I also made sure to preserve the number of consecutive newlines.

nuclearsandwich:
> There's a level that I don't think that we should do a character by character test for all the parsed outputs. But maybe just a few checks for specific things that are properly escaped etc would make sense. And we can trust that the specific behaviors are well defined by the parser that we're relying upon and we can just spec that we'll follow the parser.


> There's a level that I don't think that we should do a character by character test for all the parsed outputs. But maybe just a few checks for specific things that are properly escaped etc would make sense.

The description I put in the test is meant to be a sort of "catchall canary" for all the stuff I think we should care about such as "flattening" text values from tags, compressing whitespace according to html rendering convention, and honoring `<br>` tags as newlines in the result.

I expect this description to grow into an arcane horror as time goes on. Not the easiest thing to test individual facets but perhaps easier than separate fixtures for each element (and testing them in combination has its own benefit.)

> And we can trust that the specific behaviors are well defined by the parser that we're relying upon and we can just spec that we'll follow the parser.

XML itself makes no assertions about whitespace except that it should be handed over intact to the application which may decide what to do with it and in general should not be expected to be preserved.

From the XML specification section 2.10 on white space handling: 

> An XML processor MUST always pass all characters in a document that are not markup through to the application. A validating XML processor MUST also inform the application which of these characters constitute white space appearing in element content.

In this case, catkin_pkg is the application and should decide what to do about whitespace.
The existing `description` is being left as-is, I'm not entirely certain if the way the description is stored whether or not it is usable as an xml fragment so we may be back later to add an `xml_description` field. index.ros.org doesn't use catkin_pkg at all and is using a generic XML parser to extract the description and render it as html so that isn't a reliable indicator that the current description is correct in that case although the years without complaint means at least no one is stressing out about it.

This also means that catkin_pkg should determine how best to represent the description in plaintext contexts and I think this is a good first try. It'd be interesting to do other things like render emphasis or links using markdown style but that rapidly moves into territory that pushes the boundaries of "plain text" and leaves room for difference of opinion vs just extracting text content recursively from child nodes and doing something "sensible" with whitespace.

cottsay (approved):
This looks great, thanks for tackling it.

PR #331, opened 2022-04-12
update catkin_prepare_release to support setup.py files

wjwwood:
I know some people don't like using `catkin_prepare_release`, but I personally enjoy it because I know what magic is happening and I like the consistency it brings in commit messages, tags, and changelog headings. But until this pull request it didn't support `setup.py` based packages because it would not update the version in the `setup.py`, only the `package.xml`. This pull request changes that and adds `ament_python` to the list of supported build types for this tool.

### DIFF
--- a/src/catkin_pkg/package_version.py
+++ b/src/catkin_pkg/package_version.py
@@ -84,6 +84,29 @@ def _replace_version(package_str, new_version):
     return new_package_str
 
 
+def _replace_setup_py_version(setup_py_str, new_version):
+    """
+    Replace the version tag in contents if there is only one instance and it is
+    using a literal as the version.
+
+    :param str package_str: contents of setup.py
+    :param str new_version: new version number
+    :returns: new setup.py string
+    :rtype: str
+    :raises RuntimeError:
+    """
+    # try to replace contens
+    new_setup_py_str, number_of_subs = re.subn(
+        r'version=([\'"])\d+\.\d+\.\d+([\'"]),',
+        r'version=\g<1>%s\g<2>,' % new_version,
+        setup_py_str)
+    if number_of_subs == 0:
+        raise RuntimeError("Failed to find a normal version statement, e.g.: version='1.2.3',")
+    if number_of_subs != 1:
+        raise RuntimeError('Illegal number of version statements: %s' % (number_of_subs))
+    return new_setup_py_str
+
+
 def _check_for_version_comment(package_str, new_version):
     """
     Check if a comment is present behind the version tag and return it.
@@ -103,7 +126,8 @@ def _check_for_version_comment(package_str, new_version):
 
 def update_versions(paths, new_version):
     """
-    Bulk replace of version: searches for package.xml files directly in given folders and replaces version tag within.
+    Bulk replace of version: searches for package.xml and setup.py files directly in given folders
+    and replaces version tag within.
 
     :param list paths: folder names
     :param str new_version: version string "int.int.int"
@@ -111,6 +135,7 @@ def update_versions(paths, new_version):
     """
     files = {}
     for path in paths:
+        # Update any package.xml files.
         package_path = os.path.join(path, 'package.xml')
         with open(package_path, 'r') as f:
             package_str = f.read()
@@ -122,6 +147,17 @@ def update_versions(paths, new_version):
         except RuntimeError as rue:
             raise RuntimeError('Could not bump version number in file %s: %s' % (package_path, str(rue)))
         files[package_path] = new_package_str
+        # Update any setup.py files.
+        setup_py_path = os.path.join(path, 'setup.py')
+        if os.path.exists(setup_py_path):
+            with open(setup_py_path, 'r') as f:
+                setup_py_str = f.read()
+            try:
+                new_setup_py_str = _replace_setup_py_version(setup_py_str, new_version)
+            except RuntimeError as exc:
+                raise RuntimeError('Could not bump version number in file %s: %s' % (setup_py_path, str(exc)))
+            files[setup_py_path] = new_setup_py_str
+
     # if all replacements successful, write back modified package.xml
     for package_path, new_package_str in files.items():
         with open(package_path, 'w') as f:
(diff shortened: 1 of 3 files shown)


src/catkin_pkg/package_version.py
```diff
@@ -84,6 +84,29 @@ def _replace_version(package_str, new_version):
...
+    Replace the version tag in contents if there is only one instance and it is
+    using a literal as the version.
+
+    :param str package_str: contents of setup.py
+    :param str new_version: new version number
+    :returns: new setup.py string
+    :rtype: str
+    :raises RuntimeError:
+    """
+    # try to replace contens
```
jacobperron, line 97:
```suggestion
    # try to replace contents
```

src/catkin_pkg/package_version.py
```diff
@@ -84,6 +84,28 @@ def _replace_version(package_str, new_version):
...
+    """
+    Replace the version tag in contents if there is only one instance and it is using a literal as the version.
+
+    :param str package_str: contents of setup.py
+    :param str new_version: new version number
+    :returns: new setup.py string
+    :rtype: str
+    :raises RuntimeError:
+    """
+    # try to replace contens
```
nuclearsandwich, line 97:
```suggestion
    # try to replace contents
```

src/catkin_pkg/package_version.py
```diff
@@ -140,7 +174,7 @@ def get_forthcoming_label(rst):
             if len(section.children) > 0 and isinstance(section.children[0], docutils.nodes.title):
                 title = section.children[0]
         if title and len(title.children) > 0 and isinstance(title.children[0], docutils.nodes.Text):
-            title_text = title.children[0].rawsource
+            title_text = title.children[0].astext()
```
wjwwood, line 177: I'm not 100% sure about why this change is needed. I don't know if this was just an uncommon code path (ran into it trying this out on `launch`'s repo), but the docs agree this is the right thing to do:

http://epydoc.sourceforge.net/docutils/private/docutils.nodes.Text-class.html

And without it you get an error saying the `Text` class has no attribute `rawsource`.

src/catkin_pkg/package_version.py
```diff
@@ -122,6 +145,17 @@ def update_versions(paths, new_version):
...
         files[package_path] = new_package_str
+        # Update any setup.py files.
+        setup_py_path = os.path.join(path, 'setup.py')
+        if os.path.exists(setup_py_path):
+            with open(setup_py_path, 'r') as f:
+                setup_py_str = f.read()
+            try:
+                new_setup_py_str = _replace_setup_py_version(setup_py_str, new_version)
+            except RuntimeError as exc:
+                raise RuntimeError('Could not bump version number in file %s: %s' % (setup_py_path, str(exc)))
```
jacobperron, line 156: Is this going to raise an exception for packages that contain a setup.py without a version (e.g. a lot catkin packages in ROS 1)?

src/catkin_pkg/package_version.py
```diff
@@ -84,6 +84,28 @@ def _replace_version(package_str, new_version):
...
+    """
+    # try to replace contens
+    new_setup_py_str, number_of_subs = re.subn(
+        r'version=([\'"])\d+\.\d+\.\d+([\'"]),',
+        r'version=\g<1>%s\g<2>,' % new_version,
+        setup_py_str)
+    if number_of_subs == 0:
+        raise RuntimeError("Failed to find a normal version statement, e.g.: version='1.2.3',")
+    if number_of_subs != 1:
+        raise RuntimeError('Illegal number of version statements: %s' % (number_of_subs))
```
nuclearsandwich, line 105: I appreciate that these are hard errors. I think it's the correct thing to do and since ament_package templates use literal version strings in setup.py hopefully there aren't too many packages out there that have gotten fancy.

src/catkin_pkg/package_version.py
nuclearsandwich, line 177: Could you explain this change and its motivation?

test/test_package_version.py
```diff
@@ -45,6 +46,34 @@ def test_replace_version(self):
...
+            RuntimeError,
+            _replace_setup_py_version,
+            'version=something,',
+            '1.0.0')
+        self.assertRaises(
+            RuntimeError,
+            _replace_setup_py_version,
+            'version="0.0.1",version="0.0.1",',
+            '1.0.0')
+
```
nuclearsandwich, line 76: Thank you for adding tests! :people_hugging:

nuclearsandwich (commented):
Ah this is really promising! Left a couple of comments but this isn't far from :heavy_check_mark: as is!

src/catkin_pkg/package_version.py
nuclearsandwich: Race condition in comments. It's explained in another thread.

wjwwood: Sorry, was commenting on this one as you were reviewing: https://github.com/ros-infrastructure/catkin_pkg/pull/331#discussion_r848937009

src/catkin_pkg/package_version.py
wjwwood: Ah, that's a good point. I thought about restricting this check to the situation when your built type is `ament_python`, but that information wasn't readily available and didn't want to cut into it too much. Let me look again at it.

nuclearsandwich: That's a good point. I don't see anything which makes running the setup.py modifications conditional on the `ament_python` build type but perhaps it should be?

jacobperron (approved):
LGTM

I'm not sure how worried we are about the API change.

wjwwood:
> I'm not sure how worried we are about the API change.

Yeah, I'll defer to @cottsay and @nuclearsandwich on that.

src/catkin_pkg/package_version.py
cottsay: Related to #308.

wjwwood: Ah so this is just another instance of this?

cottsay: I **think** so, but I'm not 100% sure.

wjwwood:
Bump on this one, is there anything I need to do? Or are we just waiting on a window to merge and release?

nuclearsandwich:
> Bump on this one, is there anything I need to do? Or are we just waiting on a window to merge and release?

I had started https://github.com/ros-infrastructure/catkin_pkg/pull/332 during review but hadn't gotten around to publishing it.

nuclearsandwich (approved):
Overall I'm happy with this change.

I think the API change warrants a significant version bump and naturally, a changelog entry, although I anticipate our tools being the only consumers of the API.

tfoote:
When this is released it would be great to update the docs here: https://wiki.ros.org/action/fullsearch/bloom/Tutorials/FirstTimeRelease

PR #332, opened 2022-05-06
[Refine #331] Use package.get_build_type()

nuclearsandwich:
This is a proposal to use the existing Package.get_build_type() method. As far as I can tell in local testing this works fine and avoids adding an extra function to do the same thing.


wjwwood (approved):
This makes much more sense. Not sure why I didn't think of it. Lol

PR #337, opened 2022-05-24
Add API method for getting all build_types regardless of conditions.

nuclearsandwich:
This is more of a spike than anything. I'm not convinced this is the
best way to go for this particular issue but it was the first thought
that came to my head.

Of possibly more concern is that this gets us past the "first hurdle"
but I'm unclear why the simple version bump doesn't crash also.

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -166,6 +166,23 @@ class Package(object):
             return build_type_exports[0]
         raise InvalidPackage('Only one <build_type> element is permitted.', self.filename)
 
+    def get_build_types(self):
+        """
+        Return values of export/build_type elements without conditional filtering, or ['catkin'] if unspecified.
+
+        :returns: package build types
+        :rtype: List[str]
+        :raises: :exc:`InvalidPackage`
+        """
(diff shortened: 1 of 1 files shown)


cottsay (approved):

src/catkin_pkg/package.py
```diff
@@ -166,6 +166,18 @@ def get_build_type(self):
             return build_type_exports[0]
         raise InvalidPackage('Only one <build_type> element is permitted.', self.filename)
 
+    def get_build_types(self):
```
nuclearsandwich, line 169: @cottsay should we rename this to something like `get_unconditional_build_types()` or somesuch to make it more obvious that this is an "advanced" method?

Should we also or alternatively prefix it with `_`?
```suggestion
    def _get_unconditional_build_types(self):
```

cottsay: I don't think it would hurt, but I also think that adding docs to the function would be sufficient.

nuclearsandwich: I did decide to rename it without the `_` prefix. Documentation for the method also indicates that conditions are not considered.

PR #348, opened 2022-10-06
Fix generate_changelog when multiple tags/branches point to the same commit

orensbruli:
Signed-off-by: Esteban Martinena <orensbruli@gmail.com>


nuclearsandwich:
Do you have an example repository which exhibits this issue for testing? I think that a better experience here is necessary but I'm not sure if it makes sense to pipe the STDERR and then completely ignore it. If ignoring it is correct then it would make more sense to send it to `subprocess.DEVNULL` but if we have to handle that warning somehow we will need to expand this PR to at least check for it and echo it back.

PR #352, opened 2023-01-30
Fix AttributeError with docutils >= 0.18

### LINKED ISSUE #351: Attribute Error with docutils 0.19
Debian testing received an update of docutils from 0.17.1 to 0.19. Now this AttributeError is thrown

```  
  File "/usr/lib/python3/dist-packages/catkin_pkg/cli/tag_changelog.py", line 31, in get_forthcoming_label
    title_text = title.children[0].rawsource
AttributeError: 'Text' object has no attribute 'rawsource'. Did you mean: 'source'?
```

`rawsource` was removed in 0.18 according to the [release notes](http://www.docutils.org/RELEASE-NOTES.html#release-0-18-2021-10-26)
> Removed attributes:
>   * nodes.Text.rawsource: we store the null-escaped text in Text nodes since 0.16 so there is no additional information in the rawsource.

onjen:
Closes  #351

### DIFF
--- a/src/catkin_pkg/cli/tag_changelog.py
+++ b/src/catkin_pkg/cli/tag_changelog.py
@@ -28,7 +28,7 @@ def get_forthcoming_label(rst):
             if len(section.children) > 0 and isinstance(section.children[0], docutils.nodes.title):
                 title = section.children[0]
         if title and len(title.children) > 0 and isinstance(title.children[0], docutils.nodes.Text):
-            title_text = title.children[0].rawsource
+            title_text = title.children[0]
             if FORTHCOMING_LABEL.lower() in title_text.lower():
                 if forthcoming_label:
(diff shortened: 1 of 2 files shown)


cottsay:
Is this change backwards compatible with older docutils versions?

onjen:
I just tested with `docutils==0.16.0` and it works.

Also the [release notes](http://www.docutils.org/RELEASE-NOTES.html#release-0-18-2021-10-26) state that the text is already in the node and there is no need to access the `rawsource` attribute.

> nodes.Text.rawsource: we store the null-escaped text in Text nodes since 0.16 so there is no additional information in the rawsource.

cottsay (approved):
Manually tested with some older docutils versions and added tests to prove it works with the newer ones as well.

PR #354, opened 2023-04-25
Drop support for Python 2

cottsay:
* The minimum Python version is now 3.6
* Shebangs now use python3 specifically
* EOL Debian/Ubuntu suites have been dropped
* Upcoming Debian Bookworm has been added
* Use of the 'mock' backport package has been removed
* GitHub Actions have been updated to the latest major
* A new flake8 violation was addressed

This change should be immediately followed by a major version bump and release, leaving us plenty of room to backport fixes and features for Python 2 if we find that to be necessary someday.

### DIFF
--- a/bin/catkin_create_pkg
+++ b/bin/catkin_create_pkg
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import sys
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,20 +9,18 @@ jobs:
     build:
       strategy:
         matrix:
-          os: [ubuntu-latest, macos-latest]
+          os: [ubuntu-22.04, macos-latest]
           python: ['3.7', '3.8', '3.9', '3.10']
           include:
-          - os: ubuntu-18.04
-            python: '2.7'
-          - os: ubuntu-18.04
+          - os: ubuntu-20.04
             python: '3.6'
       name: catkin_pkg tests
       runs-on: ${{matrix.os}}
 
       steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up Python ${{matrix.python}}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{matrix.python}}
       - name: Install dependencies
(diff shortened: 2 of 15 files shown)


nuclearsandwich:
> * Upcoming Debian Bookworm

Upcoming no longer! Bookworm released on June 10 https://www.debian.org/releases/bookworm/

> This change should be immediately followed by a major version bump and release, leaving us plenty of room to backport fixes and features for Python 2 if we find that to be necessary someday.

Just to clarify, are you thinking 1.0.0 or 0.6.0? I would love to start moving infra packages into a 1.0 commitment level or even just drop the leading 0 and go directly to 6.0.

bin/catkin_create_pkg
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
nuclearsandwich, line 1: It is unclear to me if this is the right / recommended thing to do (I'm looking primarily at https://peps.python.org/pep-0394/#for-python-script-publishers) which includes

> For scripts that are only expected to be run in an activated virtual environment, shebang lines can be written as `#!/usr/bin/env python`, as this instructs the script to respect the active virtual environment.
> In cases where the script is expected to be executed outside virtual environments, developers will need to be aware of the following discrepancies across platforms and installation methods:
> 
> * Older Linux distributions will provide a `python` command that refers to Python 2, and will likely not provide a `python2` command.
> * Some newer Linux distributions will provide a `python` command that refers to Python 3.
> * Some Linux distributions will not provide a `python` command at all by default, but will provide a `python3` command by default.

I think it depends on which platforms we expect these scripts to be run. I have pretty much talked myself into preferring the `python3` since it very explicitly breaks python2 support which we are very explicitly breaking.

IMO we have latitude here to do either since the "most right" answer is probably to use entry points which will install the scripts appropriately but we're not making that change today.

nuclearsandwich (approved):
Thanks for getting this giant snowball rolling!

cottsay:
> Just to clarify, are you thinking 1.0.0 or 0.6.0? I would love to start moving infra packages into a 1.0 commitment level or even just drop the leading 0 and go directly to 6.0.

I'm indifferent. Then again, it's a good excuse to finally bump to 1.0 and start holding ourselves to semver. Let's plan on that.

bin/catkin_create_pkg
cottsay: I'm pretty sure that console_scripts entry points are installing the executables we see in the packages, but I think the scripts are still used when running tests.

I don't feel strongly. Throw me a :+1: on this comment and I'll revert it.

nuclearsandwich (approved):

PR #365, opened 2024-04-01
Fix symlink resolution on Windows w/short filenames

cottsay:
Two issues here:
1. os.path.realpath() doesn't resolve symlinks at all prior to Python 3.8. Best we can do is skip the test there.
2. It doesn't appear that os.path.realpath() is idempotent on Windows when path resolution requires both symlink resolution and 8.3/short filename resolution. Calling it twice seems to give the result we want.

### DIFF
--- a/src/catkin_pkg/workspaces.py
+++ b/src/catkin_pkg/workspaces.py
@@ -92,6 +92,9 @@ def order_paths(paths_to_order, prefix_paths):
 
 
 def _is_equal_or_in_parents(dir_, path):
+    # On Windows, when a symlink points to a path which uses 8.3/short filenames, it doesn't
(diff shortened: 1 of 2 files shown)


tfoote (approved):
That's a really weird behavior. Can you add a comment in the code with a link to a reference why it's being duplicated so that in the future a reviewer doesn't accidentally cause a regression trying to clean it up.

cottsay:
> That's a really weird behavior.

It was pretty hard to track down since I wasn't hitting the problem locally. It appears that the GitHub Actions runners use 8.3 filenames in `TMP` (probably in an effort to make the directory path shorter), which is the only reason I became aware of the misbehavior.

> Can you add a comment

Done in 7cc7ad6.

PR #371, opened 2024-08-30
Print an error level message if package does not start with a letter

nbbrooks:
While doing some adversarial testing, we found that we could attempt to create a ROS 2 C++ package that starts with a letter and ament would print a warning level message (I believe using catkin's package naming checks). There would be subsequent  compilation errors of course, but it would be good to flag this as an Error earlier if possible.

```
WARNING: Package name "1test" does not follow the naming conventions. It should start with a lower case letter and only contain lower case letters, digits, underscores, and dashes.
```

Is there a reason the error level check allows for starting with a number?

### DIFF
--- a/src/catkin_pkg/package.py
+++ b/src/catkin_pkg/package.py
@@ -243,7 +243,7 @@ class Package(object):
         if not self.name:
             errors.append('Package name must not be empty')
         # accepting upper case letters and hyphens only for backward compatibility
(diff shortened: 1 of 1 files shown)


cottsay (approved):

cottsay:
Sorry, I ended up reverting this change. It looks like there are tests asserting the current behavior that would need to be updated, but their presence makes me believe that this behavior is intentionally a "warning" and not an "error".

PR #374, opened 2025-07-03
Replace pkg_resources usage with packaging.

nuclearsandwich:
pkg_resources is deprecated and will be removed in setuptools 81 as early as 2025-11-30. Support for Python's standardized version parsing is available in the packaging module which is reliably available on our target platforms although sadly not in the standard library.

I believe that this dependency is rightly placed on the modules sub-package not the cli commands subpackage because it is possible to use the changelog interfaces through the Python API. I believe the lack of a pkg_resources / setuptools dependency on module subpackage was an oversight that was often masked by the ubiquity of setuptools in most installations.


setup.py
```diff
@@ -45,6 +45,7 @@
     'license': 'BSD',
     'install_requires': [
         'docutils',
+        'packaging',
         'python-dateutil',
         'pyparsing',
         'setuptools',
```
dotlambda, line 51:
```suggestion
```

PR #375, opened 2025-07-03
Remove spurious global declaration.

nuclearsandwich:
_ansi is not currently used as a global (it is a module variable in this file only and flake8 now catches and lints this unneeded declaration.

### DIFF
--- a/src/catkin_pkg/terminal_color.py
+++ b/src/catkin_pkg/terminal_color.py
@@ -42,13 +42,11 @@ _ansi = {}
 
 def ansi(key):
     """Return the escape sequence for a given ansi color key."""
-    global _ansi
     return _ansi[key]
 
 
 def enable_ANSI_colors():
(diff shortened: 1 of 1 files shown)


cottsay (approved):

src/catkin_pkg/terminal_color.py
```diff
@@ -42,26 +42,25 @@
 
 def ansi(key):
     """Return the escape sequence for a given ansi color key."""
-    global _ansi
     return _ansi[key]
 
 
 def enable_ANSI_colors():
     """Populate the global module dictionary `ansi` with ANSI escape sequences."""
-    global _ansi
```
nuclearsandwich, line 51: This is in the commit history now but as it happens this usage was *not* spurious nor flagged by flake8 but I assumed that 2/3 == 3/3 and removed it anyway. In order to fix the failing test without restoring the use of the global declaration I changed the code to update rather than assign and I think it's cleaner.

